PR #3, opened 2023-03-30
Updating License from MIT to MPL v2 w HCD, without Copyright

gracepotma:
This React Form Engine, now called the "OpenMRS Form Engine", previously-called the OHRI Form Engine, had an MIT license and UCSF (c). Today, with support from UCSF team members, we have worked together (OMRS and UCSF) to move the source repo into the OpenMRS community. I would like to follow our conventions and change this to MPL v2 with Healthcare Disclaimer to follow community conventions so this can be safely re-used and co-maintained by our global OpenMRS community.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,21 +1,401 @@
-MIT License
+Mozilla Public License, version 2.0
 
-Copyright (c) 2021 UCSF OHRI team
+1. Definitions
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


samuelmale (approved):
Thanks Grace!

ebambo:
@gracepotma agreed and thanks for catching that. Do you think we should apply this to all our repos since at the end they are all supposed to be used by the wider community?

gracepotma:
That would actually be very perfect @ebambo !!

MwaririM:
@gracepotma Thank you for this.

PR #4, opened 2023-04-01
(feat) Add ability to retrieve encounter type and default to `unknown` encounter role

donaldkibet:
### What does this PR do?

At the moment the `hard` code value for encounterRole is `Clinician` for some implementors this might be missing. I have added the ability to retrieve the `encounterRole` and add preference to `Clinician` as the default value if present. In its absence the `UnKnown` encounterRole is picked.

@denniskigen @samuelmale @pirupius @larslemos What are your thoughts on this approach.

### DIFF
--- a/src/hooks/useEncounterRole.tsx
+++ b/src/hooks/useEncounterRole.tsx
@@ -1,0 +1,15 @@
+import { OpenmrsResource, openmrsFetch } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr/immutable';
+
+export function useEncounterRole() {
+  const { data, error, isLoading } = useSWRImmutable<{ data: { results: Array<OpenmrsResource> } }, Error>(
+    '/ws/rest/v1/encounterrole?v=custom:(uuid,display,name)',
+    openmrsFetch,
+  );
+  const clinicalEncounterRole = data?.data.results.find(encounterRole => encounterRole.name === 'Clinician');
+
+  if (clinicalEncounterRole) {
+    return { encounterRole: clinicalEncounterRole, error, isLoading };
+  }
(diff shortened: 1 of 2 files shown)


src/hooks/useEncounterRole.tsx
```diff
@@ -0,0 +1,15 @@
+import { OpenmrsResource, openmrsFetch } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr/immutable';
+
+export function useEncounterRole() {
+  const { data, error, isLoading } = useSWRImmutable<{ data: { results: Array<OpenmrsResource> } }, Error>(
+    '/ws/rest/v1/encounterrole?v=full',
+    openmrsFetch,
+  );
+  const clinicalEncounterRole = data?.data.results.find(encounterRole => encounterRole.display === 'Clinician');
```
samuelmale, line 9: I could be wrong, but hypothetically the `display` property is a translatable string by the backend; In such scenarios, the above find operation will be broken. I would use the name property.

src/hooks/useEncounterRole.tsx
```diff
@@ -0,0 +1,15 @@
+import { OpenmrsResource, openmrsFetch } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr/immutable';
+
+export function useEncounterRole() {
+  const { data, error, isLoading } = useSWRImmutable<{ data: { results: Array<OpenmrsResource> } }, Error>(
+    '/ws/rest/v1/encounterrole?v=full',
```
samuelmale, line 6: Since we aren't doing much with the encounter role, do we need the full representation?

samuelmale (approved):
Nice one Donald, just a few minor comments

pirupius (approved):

PR #30, opened 2023-04-18
O3-2061: Introduce visit as an optional parameter to the form-engine

CynthiaKamau:
… fields

## Requirements

## Summary

 Passed active visit uuid in form if it exists

## Screenshots

visit uuid

### DIFF
--- a/src/components/encounter/ohri-encounter-form.tsx
+++ b/src/components/encounter/ohri-encounter-form.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import { SessionLocation, useLayoutType } from '@openmrs/esm-framework';
+import { SessionLocation, useLayoutType, Visit } from '@openmrs/esm-framework';
 import { ConceptFalse, ConceptTrue } from '../../constants';
 import { OHRIFormContext } from '../../ohri-form-context';
 import { getHandler, getValidator } from '../../registry/registry';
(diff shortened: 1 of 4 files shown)


src/components/encounter/ohri-encounter-form.tsx
```diff
@@ -35,6 +35,7 @@ interface OHRIEncounterFormProps {
   encounterDate: Date;
   provider: string;
   location: { uuid: string; name: string };
+  visitUUID: string;
```
samuelmale, line 38: Do you want to make this an optional prop?

samuelmale (commented):
> Pass active visit attributes as default

Changes in this PR don't fully satisfy the above. I would expect you to fetch the current visit and use that to initialise the necessary properties ie. `encounterDatetime`, `provider` and `location`

src/components/encounter/ohri-encounter-form.tsx
samuelmale: This is supposed to be an optional param. On a side note, since we may need to pick a few props from the visit, why don't we just pass the current visit object? We are most likely to have a visit object within the workspace context, that way the engine won't have to refetch it. Thought?? @donaldkibet, @denniskigen

CynthiaKamau: i can pass the whole visit

donaldkibet (approved):

samuelmale (approved):

PR #32, opened 2023-04-19
O3-2055 introduces encounter provider component to support passing encounter providers

pirupius:
## Requirements

## Summary
This PR is a first in a series to add Visit support to react form engine. This PR introduces an `EncounterProvider` component in the addition to the already existing `EncounterLocation` component. 

Next steps involves passing visit location and provider as default values and adding handlers for form submission. 

## Related Issue

https://issues.openmrs.org/browse/O3-2055

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -243,11 +243,11 @@ export interface OpenmrsFormResource extends OpenmrsResource {
   valueReference: string;
 }
 
-export interface ProviderResponse {
-  results: Array<Provider>;
+export interface EncounterProviderResponse {
+  results: Array<EncounterProvider>;
 }
 
-export interface Provider {
+export interface EncounterProvider {
   uuid: string;
   display: string;
   person: {
@@ -295,4 +295,15 @@ export interface Location {
   ];
   retired: boolean;
   attributes: [];
+}
+
+export interface DataSource<T> {
+  fetchData(): any; // Should this be a promise?
+  resolveSelectedValue?: (value: any) => any;
+  searchOptions?: (searchText: string) => Promise<Array<any>>;
+}
+
+export interface DataSourceItem {
+  uuid: string;
+  display: string;
 }
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -298,12 +298,17 @@ export interface Location {
 }
 
 export interface DataSource<T> {
+  /**
+   * Fetches arbitrary data from a data source
+   */
   fetchData(): Promise<Array<T>>;
-  makeSelectItems(): Array<UISelectItem>;
-  searchOptions?: (searchText: string) => Promise<Array<any>>;
+  /**
+   * Maps a data source item to an object with a uuid and display property
+   */
+  toUuidAndDisplay(item: T): UuidAndDisplay;
 }
 
-export interface UISelectItem {
-  id: string;
+export interface UuidAndDisplay {
+  uuid: string;
   display: string;
 }
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -191,6 +191,7 @@ export interface PostSubmissionAction {
     sessionMode: SessionMode;
   }): void;
 }
+
 // OpenMRS Type Definitions
 export interface OpenmrsEncounter {
   uuid?: string;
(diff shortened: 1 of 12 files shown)


src/api/types.ts
```diff
@@ -233,3 +234,57 @@ export interface OpenmrsFormResource extends OpenmrsResource {
...
+    display: string;
+  };
+  childLocations: [
+    {
+      uuid: string;
+      display: string;
+    },
+  ];
+  retired: boolean;
+  attributes: [];
```
pirupius, line 289: Not sure if we need all these attributes, if not then i can make adjustments to the endpoint only to return the necessary fields

src/components/encounter/ohri-encounter-form.tsx
```diff
@@ -29,6 +29,9 @@ import { useEncounter } from '../../hooks/useEncounter';
 import { useInitialValues } from '../../hooks/useInitialValues';
 import { useEncounterRole } from '../../hooks/useEncounterRole';
 
+/**
+ * Todo pirupius - this needs to be refactored to pass the provider and location as props
+ */
```
pirupius, line 34: This will build on top of work being done 👉🏾  https://github.com/openmrs/openmrs-form-engine-lib/pull/30

src/components/inputs/providers/encounter-provider.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useProviders } from '../../../api/patient-visits.resource';
+import { OHRIFormContext } from '../../../ohri-form-context';
+import { getConceptNameAndUUID } from '../../../utils/ohri-form-helper';
+import { OHRIFieldValueView } from '../../value/view/ohri-field-value-view.component';
+import { isTrue } from '../../../utils/boolean-utils';
+
+interface EncounterProviderProps {
+  question: OHRIFormField;
+  defaultValue?: any;
+  onChange?: any; // TODO - this might need to be deprecated
```
pirupius, line 15: Not sure of a use case of this but leaving this for now

samuelmale:
Some thoughts: 

What do you think of approaching this in a more generic way? We just need to separate concerns; there are three tiers here:

1. Rendering of the visual component
2. Fetching of the search-able data (combo box data list)
3. Linking the selected provider to the encounter

> Rendering of the visual component

We can develop a generic search-able component (combo-box like). This special control would expect a `DataSource` and a submission handler or what we commonly call the `type`.

> Fetching of the search-able data (combo box data list)

We can introduce an interface which all data sources will implement:

```ts

interface DataSource<T> {
  fetchData(...): Promise<T[]>;
  toComboItem(item: T): ComboItem;
}

interface ComboItem {
  id: string;
  display: string;
}

// Eg.

class EncounterProviderDataSource implements DataSource<EncounterProvider> {

   private readonly apiUrl = "v1/../provider";

    fetchData() {
      // do fetch list of providers
      return openmrsFetch(apiUrl);
    }
 
    toComboItem(provider: EncounterProvider): ComboItem {
       return { id: provider.uuid, display: provider.display };
    }
}

```

> Linking the selected provider to the encounter

We add a submission handler that will handle the data binding, just like other supported types eg. `obs` [here](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/registry/registry.ts#L136-L140)

That way we are able to reuse this component for other similar use-cases. I'm guessing Ampath's ui-extended or rather search has a generic implementation. Thoughts @ibacher, @denniskigen, @donaldkibet @pirupius

ibacher:
I haven't looked at this repo in any detail. I agree that separating the data-fetching from the rendering makes sense (we have a datasource for encounter providers in the Angular app too). I do think that `toComboItem()` pushes implementation details into the data source that don't belong there, though the general idea of a function that maps a provider into a known structure is probably a good one (`toComboItem()` sort of implies you might have a `toCheckboxItem()` or `toTextItem()`).

In Angular, most data sources implement an interface that's something like this:

```ts
interface DataSource<T = any> {
  resolveSelectedValue: (string) => T,  // uuid -> object
  searchOptions: (string) => Array<T> // query -> array of objects
}
```

I'm not saying that that's necessarily the interface to use, but it does do a decent job of both being generic enough and keeping the data sources focused on supplying data. We could even make this more "generic" in the sense that `DataSource<EncounterProvider>` seems to already imply that something consuming this is expecting an `EncounterProvider` object, but really, it probably needs an object like this:

```ts
interface DataSourceItem {
  uuid: string,
  display: string,
}
```

I.e., the view component probably shouldn't care what the `DataSourceItem` actually represents as long as it has an identifier and a display string for it.

src/api/types.ts
```diff
@@ -241,3 +242,73 @@ export interface OpenmrsFormResource extends OpenmrsResource {
...
+   * Fetches arbitrary data from a data source
+   */
+  fetchData(): Promise<Array<T>>;
+  /**
+   * Maps a data source item to an object with a uuid and display property
+   */
+  toUuidAndDisplay(item: T): UuidAndDisplay;
+}
+
+export interface UuidAndDisplay {
```
donaldkibet, line 311: Seems similar to `OpenmrsResource` exported by `@openmrs/esm-framework`

donaldkibet (commented):
@pirupius, we kindly request that you continue working on this task to help move it forward. Your efforts are greatly appreciated, and we look forward to seeing the progress you make. Thank you!

pirupius:
This work has been picked up by @kajambiya and moved to https://github.com/openmrs/openmrs-form-engine-lib/pull/85

PR #80, opened 2023-05-15
(feat) Don't show an error toast when a patient UUID is not present

denniskigen:
## Requirements

## Summary

When the user launches the form engine in a context where a patient UUID is not provided, an error toast gets displayed on the screen. An example is the form builder, where the form engine renders the provided schema in the form preview UI. Because we don't have a `patientUUID` in that context (compared to somewhere like the patient chart), the user gets shown a toast error message. Given that the utility of the patient UUID seems limited to just rendering a patient banner, it's best not to display the patient banner when a patient UUID is not present.

### DIFF
--- a/src/hooks/usePatientData.tsx
+++ b/src/hooks/usePatientData.tsx
@@ -14,19 +14,21 @@ function calculateAge(birthDate: Date): number {
   }
 }
 
-const patientGenderMap = {
-  female: 'F',
-  male: 'M',
-  other: 'O',
-  unknown: 'U',
-};
+export const usePatientData = (patientUuid: string) => {
+  const patientGenderMap = {
+    female: 'F',
+    male: 'M',
+    other: 'O',
+    unknown: 'U',
+  };
 
-export const usePatientData = patientUuid => {
   const { patient, isLoading: isLoadingPatient, error: patientError } = usePatient(patientUuid);
+
   if (patient && !isLoadingPatient) {
     // This is to support backward compatibility with the AMPATH JSON format
     patient['age'] = calculateAge(new Date(patient?.birthDate));
     patient['sex'] = patientGenderMap[patient.gender] ?? 'U';
   }
+
   return { patient, isLoadingPatient, patientError };
 };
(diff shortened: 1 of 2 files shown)


src/ohri-form.component.tsx
```diff
@@ -161,10 +170,6 @@ const OHRIForm: React.FC<OHRIFormProps> = ({
     reportError(formError, t);
   }, [formError, t]);
 
-  useEffect(() => {
-    reportError(patientError, t);
-  }, [patientError, t]);
```
denniskigen, line 166: This bit is what gets the error to appear.

src/ohri-form.component.tsx
```diff
@@ -91,7 +91,7 @@ const OHRIForm: React.FC<OHRIFormProps> = ({
   const session = useSession();
   const currentProvider = session?.currentProvider?.uuid ? session.currentProvider.uuid : null;
   const location = session && !(encounterUUID || encounterUuid) ? session?.sessionLocation : null;
-  const { patient, isLoadingPatient: isLoadingPatient, patientError: patientError } = usePatientData(patientUUID);
```
denniskigen, line 94: Removes a bunch of unnecessary aliases.

src/hooks/usePatientData.tsx
```diff
@@ -14,19 +14,21 @@ function calculateAge(birthDate: Date): number {
   }
 }
 
-const patientGenderMap = {
-  female: 'F',
-  male: 'M',
-  other: 'O',
-  unknown: 'U',
-};
```
denniskigen, line 22: This variable doesn't seem to be used anywhere else, so I've moved it into the `usePatientData` hook body.

denniskigen:
Absolutely right, @donaldkibet. Getting rid of the toast notification is the wrong approach. One easy fix to this is to pass an empty string as the patientUUID from the patient banner, and then possibly look into making the `patientUUID` an optional parameter so that it works OK in contexts where a patient is not available (such as the form builder).

samuelmale:
(Not to be taken very critically, just throwing my thoughts out here)

> and then possibly look into making the patientUUID an optional parameter so that it works OK in contexts where a patient is not available

I am not inclined to the idea of making the patientUUID an optional parameter because, upon reflection, the patient is the focal point. I understand that you are attempting to address the issue in the form-builder, where we lose reference to the patient after subsequent server resets. In my opinion, the best approach to resolving this problem is by loading "demo" patient(s) while in non-production environments. In a typical production-like environment, such a use-case may not arise. With that being said, I believe it is ideal to retain the error notification when the engine fails to load the patient, but the form should not break under such circumstances.

PR #96, opened 2023-07-03
The addDaysToDate function should not modify the original date.

Herobiam:
The problem with the function was that it directly modified the original date object using the setDate method. As a result, it caused unexpected behavior where the modification affected the original object. This led to the visit date and the next visit date being the same after adding a certain number of days to the visit date, as shown in the screenshot bellow:

Screenshot 2023-07-08 at 8 35 23 AM

To fix this issue, I created a new Date object based on the original date and perform the calculations on the new object.

Screenshot 2023-07-08 at 8 32 29 AM

Additionally, I included a check to verify if the resulting date is valid by checking if newDate.getTime() returns NaN. If the date is invalid, the function returns undefined and this was being displayed on the form.

### DIFF
--- a/src/utils/common-expression-helpers.ts
+++ b/src/utils/common-expression-helpers.ts
@@ -80,10 +80,9 @@ export class CommonExpressionHelpers {
 
   addDaysToDate = (date, days) => {
     const newDate = new Date(date);
-    dayjs(newDate).format('DD/MM/YYYY')
     newDate.setDate(newDate.getDate() + days); 
     if(isNaN(newDate.getTime())){
-      return;
+      return null;
     }
     return newDate;
   };
--- a/src/utils/common-expression-helpers.ts
+++ b/src/utils/common-expression-helpers.ts
@@ -78,13 +78,10 @@ export class CommonExpressionHelpers {
     return date;
   };
 
-  addDaysToDate = (date, days) => {
-    const newDate = new Date(date);
-    newDate.setDate(newDate.getDate() + days); 
-    if(isNaN(newDate.getTime())){
-      return null;
-    }
-    return newDate;
+  addDaysToDate = (date: Date, days: number): Date => {
+    return dayjs(date)
+      .add(days, 'day')
+      .toDate();
   };
 
   useFieldValue = (questionId: string) => {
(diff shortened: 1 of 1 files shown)


ODORA0 (changes requested):

src/utils/common-expression-helpers.ts
```diff
@@ -79,9 +79,17 @@ export class CommonExpressionHelpers {
...
+    const newDate = new Date(date);
+    dayjs(newDate).format('DD/MM/YYYY')
+    newDate.setDate(newDate.getDate() + days); 
+    if(isNaN(newDate.getTime())){
+      return;
+    }
+    return newDate;
 
-    return date;
+    // date.setDate(date.getDate() + days);
```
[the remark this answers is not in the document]
Herobiam: okay great will do so

ODORA0 (approved):

src/utils/common-expression-helpers.ts
```diff
@@ -79,9 +79,13 @@ export class CommonExpressionHelpers {
   };
 
   addDaysToDate = (date, days) => {
-    date.setDate(date.getDate() + days);
-
-    return date;
+    const newDate = new Date(date);
+    dayjs(newDate).format('DD/MM/YYYY')
```
pirupius, line 83: Shouldn't this format be based off locale?

src/utils/common-expression-helpers.ts
```diff
@@ -79,9 +79,13 @@ export class CommonExpressionHelpers {
...
+    newDate.setDate(newDate.getDate() + days);
```
donaldkibet, line 84: Since we are already using dayjs library could we use this

```
const newDate = dayjs(date).add(days, 'day');
return newDate.format('DD/MM/YYYY');
```

Herobiam: The issue with the function was that it modifies the original date object directly using the setDate method. This lead to unexpected behavior because the modification affects the original object.

To fix this issue, you should create a new Date object based on the original date and perform the calculations on the new object.

To ensure proper calculations on the date I used the dayjs library to format the date in the "DD/MM/YYYY" format, and adding the specified number of days to the new date. Additionally, I included a check to verify if the resulting date is valid by checking if newDate.getTime() returns NaN. If the date is invalid, the function returns undefined and this was being displayed on the form, @samuelmale if this is not clear then we can have a session to explain better with an example

samuelmale (approved):

PR #107, opened 2023-08-17
Moved schema-validator logic to form engine

arodidev:
Moves the form-schema validator logic to live inside the form-engine


src/schema-validator.ts
```diff
@@ -0,0 +1,139 @@
...
+        .join(',')
+    : '';
+
+  if (searchRef) {
+    try {
+      const { data } = await openmrsFetch(`/ws/rest/v1/concept?references=${searchRef}&v=${conceptRepresentation}`);
+      if (data.results.length) {
+        const [resObject] = data.results;
+
+        resObject.datatype.name === 'Boolean' &&
```
samuelmale, line 57: Should be driven by a config

src/schema-validator.ts
```diff
@@ -0,0 +1,139 @@
...
+    ? conceptObject.questionOptions.conceptMappings
+        ?.map(mapping => {
+          return `${mapping.type}:${mapping.value}`;
+        })
+        .join(',')
+    : '';
+
+  if (searchRef) {
+    try {
+      const { data } = await openmrsFetch(`/ws/rest/v1/concept?references=${searchRef}&v=${conceptRepresentation}`);
```
samuelmale, line 53: I'm concerned about the fact that we have fused the validation logic with API-specific logic. I advise reusing the existing `useConcepts()` hook. This means the validator becomes a React hook:

```js

// the validator runs if doValidate is set to true
// isValidating shows the validation status
const { errors, warnings, isValidating } = useSchemaValidationResults(doValidate, schema);

```

Alternatively, you can define a `FormSchemaValidator` interface but ensure you fetch the concepts in one API hit with proper separation of concerns.

samuelmale (changes requested):
Move the validator to https://github.com/openmrs/openmrs-form-engine-lib/tree/main/src/validators

src/components/inputs/date/ohri-date.component.tsx
```diff
@@ -169,6 +169,7 @@ const OHRIDate: React.FC<OHRIFormFieldProps> = ({ question, onChange, handler })
                 invalid={!isFieldRequiredError && errors.length > 0}
                 invalidText={errors[0]?.message}
                 warn={warnings.length > 0}
+                readOnly={question.readonly}
```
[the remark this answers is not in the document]
arodidev: Apologies, there was a mix up of the branches I was working with, resolving this now

src/components/inputs/date/ohri-date.component.tsx
```diff
@@ -180,6 +181,7 @@ const OHRIDate: React.FC<OHRIFormFieldProps> = ({ question, onChange, handler })
               id={question.id}
               labelText="Time:"
               placeholder="HH:MM"
+              readOnly={question.readonly}
```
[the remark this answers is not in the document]
arodidev: Same case here

src/form-validator.ts
```diff
@@ -0,0 +1,137 @@
+import { openmrsFetch } from '@openmrs/esm-framework';
+
+export const handleFormValidation = async (schema, configObject) => {
+  const errorsArray = [];
+  const warningsArray = [];
```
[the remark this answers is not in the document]
arodidev: Done.

src/schema-validator.ts
arodidev: @samuelmale kindly expound on this

src/validators/schema-validator.ts
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      const { data } = await openmrsFetch(`/ws/rest/v1/concept?references=${searchRef}&v=${conceptRepresentation}`);
+      if (data.results.length) {
+        const [resObject] = data.results;
+
+        resObject.datatype.name === 'Boolean' &&
+          conceptObject.questionOptions.answers.forEach(answer => {
+            if (
+              answer.concept !== 'cf82933b-3f3f-45e7-a5ab-5d31aaee3da3' &&
+              answer.concept !== '488b58ff-64f5-4f8a-8979-fa79940b1594'
```
samuelmale, line 61: I would do something like:

```js

if ( ![ConceptTrue, ConceptFalse].includes(answer.concept) ) {
   // handle error
}

```

The ConceptTrue and ConceptFalse are driven by a System configuration: `concept.true` and `concept.false`. We need to be picking these values from the backend. On the contrary, these are currently [defined as constants](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/constants.ts#L1-L2) within the library which isn't the conventional practice. I've always wanted to address this technical debt but never got the latitude. 

Ideally, we should introduce a new hook that will read this system setting. It would be great if you offered to address this otherwise you can use the already defined constants and we address this later on.

src/schema-validator.ts
samuelmale: My bad @arodidev my comment regarding the config, [see](https://github.com/openmrs/openmrs-form-engine-lib/pull/107/files#r1302767224). 

As a good practice, I would create an `enum` holding all the know datatypes to prevent hard-coding these.

```JS

const ConceptDataTypes {
  Boolean: 'Boolean',
  Coded: 'Coded',
  Numeric: 'Numeric',
  Text: 'Text'
  // more...
};


// validator.js


if (concept.datatype.name === ConceptDataTypes.Boolean) {}



```

src/schema-validator.ts
arodidev: @samuelmale I had a previous implementation of how I understood your statement [here](https://github.com/arodidev/openmrs-esm-form-builder/blob/concept-mappings/src/components/interactive-builder/interactive-builder.component.tsx) , kindly advise if this logic suffices.

src/validators/schema-validator.ts
arodidev: This, is well noted. I would also like to take on the issue of the hook reading the values from the backend if possible.

src/hooks/useSchemaValidator.ts
```diff
@@ -0,0 +1,218 @@
+import { useDatatype } from './useDatatype';
+import { useEffect, useMemo, useState } from 'react';
+import { useConfig } from '@openmrs/esm-framework';
+import { ConceptTrue, ConceptFalse } from '../constants';
+import { OHRIFormField, OHRIFormSchema } from '../api/types';
+
+export function useSchemaValidator(schema: OHRIFormSchema) {
```
samuelmale, line 7: Two things here: 
1. We need a trigger state (to trigger the schema validation)
2. On top of the errors and warnings, it's important to return a progress state i.e.. `isValidating`

src/hooks/useSchemaValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+  //initializing state
```
samuelmale, line 8: Get rid of the comment

src/hooks/useSchemaValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+  const [errors, setErrors] = useState([]);
+  const [warnings, setWarnings] = useState([]);
+  const [fullArray, setFullArray] = useState<Array<OHRIFormField>>([]);
+  const [answersArray, setAnswersArray] = useState([]);
+  const [conceptSet, setConceptSet] = useState<Set<string>>(new Set());
+  const [answerConceptSet, setAnswerConceptSet] = useState<Set<string>>(new Set());
+  const { dataTypeToRenderingMap, conceptDataTypes } = useConfig();
+
+  const unresolvedConceptsFunc = (fullArray, filteredSetArray) => {
```
samuelmale, line 17: The signature of this function can be improved, starting with the function name. Assuming it finds unresolved concepts, `findUnresolvedConcepts` is more descriptive. 

---
`fullArray`: It's not straightforward to tell the kind of objects this variable holds same applies to the other param

src/hooks/useSchemaValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+import { useEffect, useMemo, useState } from 'react';
+import { useConfig } from '@openmrs/esm-framework';
+import { ConceptTrue, ConceptFalse } from '../constants';
+import { OHRIFormField, OHRIFormSchema } from '../api/types';
+
+export function useSchemaValidator(schema: OHRIFormSchema) {
+  //initializing state
+  const [errors, setErrors] = useState([]);
+  const [warnings, setWarnings] = useState([]);
+  const [fullArray, setFullArray] = useState<Array<OHRIFormField>>([]);
```
samuelmale, line 11: Can you rename this variable to clearly reflect that it's holding fields?

src/hooks/useDatatype.ts
```diff
@@ -0,0 +1,19 @@
+import useSWRImmutable from 'swr/immutable';
+import { openmrsFetch, OpenmrsResource } from '@openmrs/esm-framework';
+
+const conceptRepresentation =
+  'custom:(uuid,display,datatype,conceptMappings:(conceptReferenceTerm:(conceptSource:(name),code)))';
+
+export function useDatatype(references: Set<string>) {
```
samuelmale, line 7: Assuming this returns concepts, what stops you from reusing the `useConcepts()` hook?

src/api/types.ts
```diff
@@ -257,3 +260,13 @@ export interface DataSource<T> {
...
+
+export interface ConfigObject {
+  Numeric: Array<string>;
+  Coded: Array<string>;
+  Text: Array<string>;
+  Date: Array<string>;
+  Datetime: Array<string>;
+  Boolean: Array<string>;
+  Rule: Array<string>;
+}
```
samuelmale, line 272: This is possessing a generic name; one may assume it's the general config interface for the form-engine configuration yet I think it's a form schema validator configuration?

src/hooks/useSchemaValidator.ts
```diff
@@ -0,0 +1,218 @@
...
+  const [answersArray, setAnswersArray] = useState([]);
+  const [conceptSet, setConceptSet] = useState<Set<string>>(new Set());
+  const [answerConceptSet, setAnswerConceptSet] = useState<Set<string>>(new Set());
+  const { dataTypeToRenderingMap, conceptDataTypes } = useConfig();
```
samuelmale, line 15: Have you actually tested this out? I don't see where you defined the config schema plus you would have to load this config by module name:

```jsx
const { dataTypeToRenderingMap, conceptDataTypes } = useConfig({
    externalModuleName: "@openmrs/openmrs-form-engine-lib",
  });

```

src/hooks/useSchemaValidator.ts
arodidev: Noted.

src/hooks/useSchemaValidator.ts
arodidev: Noted.

src/hooks/useSchemaValidator.ts
arodidev: Noted.

src/hooks/useSchemaValidator.ts
arodidev: Noted.

src/hooks/useDatatype.ts
arodidev: I think we had this conversation a while back to create a new custom hook that also returns a filtered set with unresolved concepts, but if need be I can revert back to using the useConcept hook and modifying it.

src/api/types.ts
arodidev: Noted.

PR #112, opened 2023-08-21
(feat) Use Function() rather than eval()

ibacher:
## Requirements

## Summary

This builds on #110 and #111, mostly for convenience.

As Mozilla's docs say [never use eval()!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!). This replaces `eval()` with the use of a Function constructor.

This is potentially breaking if certain forms were relying on variables that are not part of the (much stricter) function context, but this does pass all existing tests and adding new variables to the context is not hard, simply a matter of adding them to the [expression context object](https://github.com/openmrs/openmrs-form-engine-lib/compare/feat/use-fn-not-eval?expand=1#diff-ca30996c3884289ab65379c8db179aef47a3582d20b03e93663b3c91e6565559R38) and [this one](https://github.com/openmrs/openmrs-form-engine-lib/compare/feat/use-fn-not-eval?expand=1#diff-ca30996c3884289ab65379c8db179aef47a3582d20b03e93663b3c91e6565559R80).

The reason for doing this is that code run within an `eval()` context closes over the containing scope, which means that certain operations cannot be optimised by the JavaScript runtime; it also means that the `eval()` context might, unintentionally, have access to certain objects that are in the parent scope, but not intended to be in the evaluation scope. With Function constructors, the evaluation context is now strictly limited to those values made available through the `expressionContext` variable.

### DIFF
--- a/src/hooks/useInitialValues.test.ts
+++ b/src/hooks/useInitialValues.test.ts
@@ -115,12 +115,11 @@ jest.mock('../utils/expression-runner', () => {
   };
 });
 
-describe('useInialValues', () => {
+describe('useInitialValues', () => {
   const encounterDate = new Date();
 
   afterEach(() => {
-    allFormFields = allFormFields.slice(0, 6);
-    allFormFields.forEach(field => {
+    allFormFields.slice(0, 6).forEach(field => {
       delete field.value;
     });
   });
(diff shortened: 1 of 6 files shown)


samuelmale:
Great work @ibacher! I've always fancied moving to Function based evaluations but there has always been something in the way.

PR #113, opened 2023-08-22
O3-2266 OHRI-1610 Introduce Location Datasource

kajambiya:
## Requirements

## Summary
This PR introduces the location data source. It allows locations to be captured as obs with the help of the ui-select-extended control.
The PR also include some code refactor to the data-source component.

## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/12844964/4bd1d10a-d8e5-48f9-9b49-740ac06492e1


src/components/inputs/ui-select-extended/ui-select-extended.tsx
```diff
@@ -26,8 +26,13 @@ export const UISelectExtended: React.FC<OHRIFormFieldProps> = ({ question, handl
...
   const isFieldRequiredError = useMemo(() => errors[0]?.errCode == fieldRequiredErrCode, [errors]);
   const [previousValueForReview, setPreviousValueForReview] = useState(null);
-  const inputValue = useRef('');
-  const dataSource = useMemo(() => getDataSource(question.questionOptions['datasource']), []);
+  const [inputValue, setInputValue] = useState('');
+  const [isSearching, setIsSearching] = useState(false);
+  const isProcessingSelection = useRef(false);
+
+  const dataSource = useMemo(() => getDataSource(question.questionOptions['datasource'].id), []);
+  const config = useMemo(() => question.questionOptions['datasource']?.config, []);
```
samuelmale, line 34: This is an overkill value memoization. Instead, I would do something like:

```JS

const [dataSource, config] = useMemo(() => [
  getDataSource(question.questionOptions['datasource'].id),
  question.questionOptions['datasource']?.config
], []);

```

src/components/inputs/ui-select-extended/ui-select-extended.tsx
```diff
@@ -26,8 +26,13 @@ export const UISelectExtended: React.FC<OHRIFormFieldProps> = ({ question, handl
   const [displayValue, setDisplayValue] = useState('');
   const isFieldRequiredError = useMemo(() => errors[0]?.errCode == fieldRequiredErrCode, [errors]);
   const [previousValueForReview, setPreviousValueForReview] = useState(null);
-  const inputValue = useRef('');
-  const dataSource = useMemo(() => getDataSource(question.questionOptions['datasource']), []);
+  const [inputValue, setInputValue] = useState('');
+  const [isSearching, setIsSearching] = useState(false);
```
samuelmale, line 30: Do you need this?

src/datasources/data-source.ts
```diff
@@ -0,0 +1,29 @@
...
+  private url: string;
+
+  constructor(url: string) {
+    this.initialUrl = url;
+    this.url = url;
+  }
+
+  fetchData(searchTerm: string, config?: Record<string, any>): Promise<any[]> {
+    this.url = this.initialUrl;
+    if (config?.tag) {
```
ibacher, line 15: There seems to be too many things going on here. If `tag` is set we set both the `tag` and the returned value. It seems like it would be better to just have this in an `OpenmrsLocationDataSource` class or something. (We don't actually have a `tag` concept for any other core data type).

src/datasources/data-source.ts
```diff
@@ -0,0 +1,29 @@
...
+      this.url = `${this.url}?tag=${config.tag}&v=custom:(uuid,display)`;
+    }
+    return openmrsFetch(searchTerm ? `${this.url}&q=${searchTerm}` : this.url).then(({ data }) => {
+      return data.results;
+    });
+  }
+
+  toUuidAndDisplay(data: any): OpenmrsResource {
```
ibacher, line 23: Is there any reason this isn't just:

```js
toUuidAndDisplay(data: OpenmrsResource): OpenmrsResource {
  return data;
}
```

I can maybe see it being useful for other kinds of resources.

src/api/types.ts
```diff
@@ -251,7 +251,7 @@ export interface DataSource<T> {
   /**
    * Fetches arbitrary data from a data source
    */
-  fetchData(searchTerm?: string): Promise<Array<T>>;
+  fetchData(searchTerm?: string, config?: Record<string, any>): Promise<Array<T>>;
```
ibacher, line 254: Should this be `config` or should it be `parameters` or something since it isn't really "configuring"; it's providing additional arguments.

src/components/inputs/ui-select-extended/ui-select-extended.tsx
```diff
@@ -128,10 +125,26 @@ export const UISelectExtended: React.FC<OHRIFormFieldProps> = ({ question, handl
...
+              }}
               disabled={question.disabled}
               onInputChange={value => {
-                inputValue.current = value;
+                //////
+                //The onInputChange function is run immediately after the onChange function.
+                //More often than not, at this level, the selected value is not equal to the search value
+                //which leads to a call to the debounce function.
+                //This check ensures that after a value is selected, the debounce function is not called again
+                //////
```
samuelmale, line 141: I would put the comment within the if statement:

```js
// Notes:
// When the user selects a value, both the onChange and onInputChange functions are invoked sequentially.
// Issue: onInputChange modifies the search term, unnecessarily triggering a search.
```

samuelmale (commented):
Nice work @kajambiya. Just a few minor comments

src/datasources/data-source.ts
kajambiya: The datasource module has been implemented in such a way that to define a new data source, you add an entry in the registry with an id and initialize the datasource component with a url. This should be ok for most datasources and in the json you would only need to specify:
`"datasource" : {
    "id": "datasource_id"
}`
However there are datasources that would need extra parameters such as location which has a `tag` parameter. This needs to be passed at the json because the user needs to determine which locations they want for a particular form. This is what led to the introduction of the config parameter passed in the json file

src/datasources/data-source.ts
kajambiya: For us to do this, we need to make an assumption that it will be standard practice for anyone adding a datasource url to include `v=custom:(uuid,display)`

src/datasources/data-source.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.initialUrl = url;
+    this.url = url;
+  }
+
+  fetchData(searchTerm: string, config?: Record<string, any>): Promise<any[]> {
+    this.url = this.initialUrl;
+    if (config?.tag) {
+      this.url = `${this.url}?tag=${config.tag}&v=custom:(uuid,display)`;
+    }
+    return openmrsFetch(searchTerm ? `${this.url}&q=${searchTerm}` : this.url).then(({ data }) => {
```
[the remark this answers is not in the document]
kajambiya: If I get you clearly here, I thought if `config.tag == undefined`, then  `if (config?.tag) {
    }` would evaluate to false

https://github.com/openmrs/openmrs-form-engine-lib/assets/12844964/961c22a3-fcfc-4bdc-98b4-081629b954e7

src/datasources/data-source.ts
samuelmale: This being the `BaseOpenmrsDataSource` it's a fair assumption. I agree with @ibacher

ibacher: @kajambiya my suggestion isn't introducing a _new_ assumption, i.e., this code already assumes that the `data` object has both a `uuid` and `display` fields. I don't really see how we can avoid making that assumption other than something like:

```js
if (typeof data.uuid === "undefined" || typeof data.display === "undefined") {
  throw new Error();
}
```

src/datasources/data-source.ts
ibacher: That makes sense to me. On the Angular side of things, the data source interface is less strictly-defined, but most data sources implement an interface like:

```
interface DataSource<T> {
  searchOptions: (q: string) => Array<T>;
  resolveSelectedValue: (value: string) => T
}
```

Each data source can support some degree of configuration, but that's effectively passed to the constructor for the data source rather than as an argument to the `fetchData()` function.

kajambiya: @ibacher & @samuelmale I have refactored the code to make the base implementation as generic as possible. For special cases like location which may require extra parameters, we introduce a new component.

src/datasources/data-source.ts
```diff
@@ -0,0 +1,26 @@
+import { openmrsFetch, OpenmrsResource } from '@openmrs/esm-framework';
+import { DataSource } from '../api/types';
+
+export class BaseOpenMRSDataSource implements DataSource<OpenmrsResource> {
+  private initialUrl: string;
```
samuelmale, line 5: This variable seems to be redundant

src/datasources/location-data-source.ts
```diff
@@ -0,0 +1,30 @@
+import { openmrsFetch, OpenmrsResource } from '@openmrs/esm-framework';
+import { DataSource } from '../api/types';
+
+export class LocationDataSource implements DataSource<OpenmrsResource> {
+  private initialUrl: string;
```
samuelmale, line 5: Do you need this?

src/datasources/location-data-source.ts
```diff
@@ -0,0 +1,30 @@
+import { openmrsFetch, OpenmrsResource } from '@openmrs/esm-framework';
+import { DataSource } from '../api/types';
+
+export class LocationDataSource implements DataSource<OpenmrsResource> {
```
samuelmale, line 4: Why doesn't this class extend the base data source?

```suggestion
export class LocationDataSource extends BaseOpenMRSDataSource {
```

That way you don't have to implement the `toUuidAndDisplay()`

src/datasources/location-data-source.ts
```diff
@@ -0,0 +1,30 @@
...
+      this.url = `${urlParts[0]}?tag=${config.tag}&${urlParts[1]}`;
+    }
+    return openmrsFetch(searchTerm ? `${this.url}&q=${searchTerm}` : this.url).then(({ data }) => {
+      return data.results;
+    });
+  }
+
+  toUuidAndDisplay(data: OpenmrsResource): OpenmrsResource {
+    if (typeof data.uuid === 'undefined' || typeof data.display === 'undefined') {
+      throw new Error();
```
samuelmale, line 26: This was thrown error was only necessary in the base implementation because it's a generic class.

src/datasources/data-source.ts
```diff
@@ -0,0 +1,26 @@
...
+  fetchData(searchTerm: string): Promise<any[]> {
+    this.url = this.initialUrl;
+    return openmrsFetch(searchTerm ? `${this.url}&q=${searchTerm}` : this.url).then(({ data }) => {
+      return data.results;
+    });
+  }
+
+  toUuidAndDisplay(data: OpenmrsResource): OpenmrsResource {
+    if (typeof data.uuid === 'undefined' || typeof data.display === 'undefined') {
+      throw new Error();
```
samuelmale, line 22: Can you provide an error message? Something like: 
```ts
throw new Error("'uuid' or 'display' not found in the OpenMRS object.");
```

src/datasources/location-data-source.ts
```diff
@@ -0,0 +1,30 @@
...
+  private initialUrl: string;
+  private url: string;
+
+  constructor(url: string) {
+    this.initialUrl = url;
+    this.url = url;
+  }
+
+  fetchData(searchTerm: string, config?: Record<string, any>): Promise<any[]> {
+    this.url = this.initialUrl;
```
samuelmale, line 14: Are there cases where you envision the URL to change for this Datasource? Why parameterize it? Why not fuse it within assuming this Datasource will only be used for locations? Ideally, it should accept adhoc query params through configurations and the search term but I'm not sure I understand why we require a URL in the constructor. 

The parameter only makes sense for data sources whose URL might change like the base data source.

samuelmale (approved):
LGTM, thanks @kajambiya

PR #114, opened 2023-08-23
fix/ O3-2253:Marking a field as readonly should render the field control as opposed to switching to view mode

arodidev:
## Requirements

## Summary

In "edit" and "enter" mode, the fields marked as read-only should render the controlled in readOnly mode and not the default view mode display

## Screenshots

Screenshot 2023-08-23 at 15 50 40

## Related Issue

Link to community ticket: https://issues.openmrs.org/browse/O3-2303

### DIFF
--- a/src/components/inputs/date/ohri-date.component.tsx
+++ b/src/components/inputs/date/ohri-date.component.tsx
@@ -132,7 +132,7 @@ const OHRIDate: React.FC<OHRIFormFieldProps> = ({ question, onChange, handler })
     }
   }, [field.value, time]);
 
-  return encounterContext.sessionMode == 'view' || isTrue(question.readonly) ? (
+  return encounterContext.sessionMode == 'view' ? (
     <OHRIFieldValueView
       label={question.label}
       value={field.value instanceof Date ? getDisplay(field.value, question.questionOptions.rendering) : field.value}
@@ -170,6 +170,7 @@ const OHRIDate: React.FC<OHRIFormFieldProps> = ({ question, onChange, handler })
                 invalidText={errors[0]?.message}
                 warn={warnings.length > 0}
                 warnText={warnings[0]?.message}
+                readOnly={question.readonly}
               />
             </DatePicker>
           </div>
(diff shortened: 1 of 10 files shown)


arodidev:
@samuelmale @ibacher there is an issue with the carbon radio component ignoring the readOnly attribute passed to it, perhaps it has to do with the carbon version...

ibacher:
@arodidev That's nothing to do with Carbon. The underlying HTML input element [doesn't support a `readonly` attribute for all input types](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input). It should be defined (probably in the ticket) what the behaviour for those fields should be.

samuelmale:
@ibacher I agree there is no native support for `readonly` for the radio element but the latest version of Carbon somehow [supports this](https://react.carbondesignsystem.com/?path=/story/components-radiobutton--playground). I guess it's a CCS tweak in the radio group.

samuelmale (approved):
LGTM

ibacher:
@samuelmale Yeah, I missed that when I looked at Storybook earlier. The `readOnly` attribute can be applied to the `RadioButtonGroup` component, not the `RadioButton` component, although there, they [just block `onChange()` from doing anything](https://github.com/carbon-design-system/carbon/blob/95d40f7272a71033f058a881327fee42b2ab6842/packages/react/src/components/RadioButtonGroup/RadioButtonGroup.tsx#L171) and apply some styles.

samuelmale:
>The readOnly attribute can be applied to the RadioButtonGroup component, not the RadioButton component

@ibacher you're correct, but it doesn't work with our support Carbon version.

ibacher:
@samuelmale True... but we could always just add the "do nothing" on change handler if the field is set to read-only right?

samuelmale:
@ibacher that's a brilliant idea but assuming the radio buttons remain interactive, wouldn't this confuse the users?

ibacher:
That's true. We could also add the same CSS rule that gets applied:

```scss
  @use "@carbon/colors";
  .cds--radio-button-group--readonly
    .cds--radio-button
    + .cds-radio-button__label
    .cds--radio-button__appearance {
    border-color: colors.$icon-disabled;
  }

  .cds--radio-button-group--readonly .#{$prefix}--radio-button__label {
    cursor: default;
  }

  .cds--radio-button-group--readonly
    .cds--radio-button__label-text {
    cursor: text;
    user-select: text;
  }
```

Then just add the `.cds--radio-button-group--readonly` to `RadioButtonGroup`. The BEM class structure and entity layout is the same, just the rules aren't present yet.

PR #120, opened 2023-09-18
Fixes on next visit date and treatment end date calculations (minor code change) - these fixes are required by some of ETHIOHRI forms

melkam-mekonnen:
## Summary

So my PR aims to address an issue we faced on some of ETHIOHRI forms. We have created some calculations on the form-engine about a year ago, these calculations take used input values from form fields, do some background calculation and show calculated values on other input fields. However, as we modified our concepts on OCL, 2 calculations (treatment end date and next visit date calculations) were affected by that. This PR fixes the issue, making these calculations work according to the concept modifications.

### DIFF
--- a/src/utils/common-expression-helpers.ts
+++ b/src/utils/common-expression-helpers.ts
@@ -133,21 +133,42 @@ export class CommonExpressionHelpers {
     return resultViralLoadStatus ?? null;
   };
 
-  calcNextVisitDate = (followupDate, arvDispensedInDays) => {
-    let resultNextVisitDate: Date;
-    if (followupDate && arvDispensedInDays) {
-      resultNextVisitDate = new Date(followupDate.getTime() + arvDispensedInDays * 24 * 60 * 60 * 1000);
+  dispensedDoseInNumber = (dispensedDose: string) => {
+    if (dispensedDose == 'fba421cf-a483-4329-b8b1-6a3ef16081bc') {
+      return 30;
+    } else if (dispensedDose == '75d94023-7804-44f8-9998-9d678488af3e') {
+      return 60;
+    } else if (dispensedDose == '4abbd98d-0c07-42f4-920c-7bbf0f5824dc') {
+      return 90;
+    } else if (dispensedDose == '684c450f-878b-4b96-ab1b-2b539c30f033') {
+      return 120;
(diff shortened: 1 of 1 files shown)


src/utils/common-expression-helpers.ts
```diff
@@ -133,21 +133,42 @@ export class CommonExpressionHelpers {
...
+    if (dispensedDose == 'fba421cf-a483-4329-b8b1-6a3ef16081bc') {
+      return 30;
+    } else if (dispensedDose == '75d94023-7804-44f8-9998-9d678488af3e') {
+      return 60;
+    } else if (dispensedDose == '4abbd98d-0c07-42f4-920c-7bbf0f5824dc') {
+      return 90;
+    } else if (dispensedDose == '684c450f-878b-4b96-ab1b-2b539c30f033') {
+      return 120;
+    } else if (dispensedDose == 'e5f7cc4d-922a-4838-8c75-af9bdbb59bc8') {
+      return 180;
```
samuelmale, line 146: @melkam-mekonnen Unfortunately, we can't have concepts or drugs hard-coded within the form engine because they are an implementation detail. Concept dictionaries vary across implementations; we just can't afford parity for all distributions. That being said, we have to explore other options. The most forward one is by defining this function through your custom esm and registering it with the form engine. Unfortunately, this feature isn't ready as we speak but we reprioritized it and hopefully it will be ready by the end of this week.

melkam-mekonnen: Yes, that's understandable @samuelmale. Having these functions in ETHIOHRI and registering them with the form-engine sounds like the right way to go about it. We would really appreciate it if you can prioritize this feature, as you know we have already deployed ETHIOHRI for facilities last week and these calculations are part of some of the forms these facility clerks fill-out regularly. Thank you.

ebambo: @melkam-mekonnen the PR that address this was merged please have a look [here](https://github.com/openmrs/openmrs-form-engine-lib/pull/122) and close this PR since it won't be merged. Let us know if you need any assistance.

melkam-mekonnen: Thanks @ebambo, I will look into it. Removing this PR.

PR #125, opened 2023-10-03
OHRI-1839 Support for rendering problem by concept class

kajambiya:
## Requirements

## Summary
Using the ui-select-extended control to render concepts of class Diagnosis and Symptom.


src/datasources/class-array-datasource.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  fetchData(searchTerm: string, config?: Record<string, any>): Promise<any[]> {
+    return openmrsFetch(searchTerm ? `${this.url}&q=${searchTerm}` : this.url).then(({ data }) => {
+      if(config?.class){
+        return data.results.filter(concept => concept.conceptClass && config.class.includes(concept.conceptClass.uuid))
+      }
+      return data.results;
+    });
+  }
+}
```
samuelmale, line 17: @kajambiya how is this data source different from the concept data source?

src/api/types.ts
```diff
@@ -259,3 +259,14 @@ export interface DataSource<T> {
...
+
+export interface ControlTemplate{
+  name: string,
+  datasource: DataSourceParameters
+}
+
+export interface DataSourceParameters{
+  name: string,
+  config?: Record<string, any>
+}
```
samuelmale, line 271: Can you apply linter?

src/datasources/class-array-datasource.ts
kajambiya: So I wanted concept datasource to be used for fetching either all concepts or concepts by class if one specified the class. For example, for drug, we pass the drug class in the url when calling the api and fetch all concepts of class drug.

Class array datasource (naming could be improved) on the other hand was to fetch concepts from more than one class. Since we can't pass an array of classes to the url, the logic is to first fetch all concepts and then filter out those that belong to the specified classes. That's why I separated the 2 so that each can handle the certain type of use cases.

kajambiya: Had preferred to separate concerns so that there are as minimal conditions as possible but let me make the update and you take a look

samuelmale: I agree with @ebambo on this. We can simply make the concept ds more configurable. It can accept an array of concept classes. If the length of the concept classes is > 1, we apply a filter to the search results. If it's simply one we apply the concept class as a query parameter.

kajambiya: @eudson @samuelmale 
I've refactored the code accordingly. Please review and revert.

ebambo (approved):
LGTM

PR #126, opened 2023-10-04
(feat) support for reusable form components

pirupius:
## Requirements

## Summary
This PR introduces logic for support for form components previously in the [Angular Form Engine](https://ampath-forms.vercel.app/docs/core-concepts/components) as part of the feature parity milestone. The React form engine already had support for reusing forms but differed in behaviour. 

### Commonly asked questions?

**Qn:** What's the difference between sub-forms and form components?
**Ans:** Both features are built with reuse in mind. But differ in the way they're created, referenced and use cases.
- Sub-Form: Sub Forms are what the RFE already support with a user able to reference another form. The referenced sub-form is then automatically inserted by the engine and runtime bringing with it any special functionality the subform might have as well content it has as a whole. The user isn't able to exclude questions or sections from the sub-form.

- Form Components: This on the other hand is a feature that the [angular form engine (AFE component docs)](https://ampath-forms.vercel.app/docs/core-concepts/components) already had support for. With this, the user building a form had the power to reuse either part of a form component or all sections of a component. For example if you had a form for Antenatal, you could decide to use only a particular section or questions of the `component_art` not necessarily the whole component_art form.

**Qn:** Does this impact my already existing forms with the Ampath JSON schema?
**Ans:** No. The main goal with this PR for already existing implementations is to provide a smooth transition without having to redesign their forms or at least make the migration less painful. For new implementations, this is an additional tool used to design forms with reuse in mind based on the business needs. 

## Other

A follow up PR will be made to the O3 form builder to aid faster creation and reuse of form components.


pirupius:
>What's the main difference between a Sub-Form and a Component Form? Is a Component Form specific to a situation where one is embedding a component/mini-app into a bigger form? So unlike a Sub-Form, a Component also brings it's own special functionality, as well as it's own "form" content with it?

@gracepotma Some parts of this are true and others are a little interchanged. 

**Sub-Form:** Sub Forms are what is already available with a user able to reference another form. The referenced sub-form is then automatically inserted by the engine and runtime bringing with it any special functionality the subform might have as well content it has as a whole. 
In summary, it inserts the subform as is.


**Form Components:**  This on the other hand is a feature that the [angular form engine (AFE component docs)](https://ampath-forms.vercel.app/docs/core-concepts/components) supported and with this, the person building a form had the power to reuse either part of a component or all sections of a component. For example if you had a form for Antenatal, you could decide to use only a particular section or questions of the `component_art` not necessarily the whole `component_art` form. 

Both features are built with reuse in mind. But differ in the way they're created, referenced and use cases.

.github/workflows/ci.yml
```diff
@@ -174,37 +174,37 @@ jobs:
...
+
+  #     - name: Create PR
+  #       uses: peter-evans/create-pull-request@v3
+  #       with:
+          # token: ${{ secrets.ACTIONS_OHRI_PAT }}
+  #         commit-message: "(chore) Bump `@openmrs/openmrs-form-engine-lib`"
+  #         title: "(chore) Bump `@openmrs/openmrs-form-engine-lib`"
+  #         body: "This is an automated PR that bumps the [form engine library](https://github.com/openmrs/openmrs-form-engine-lib) to the newest available version tagged `next`."
+  #         branch: chore/bump-form-engine
+  #         base: main
```
samuelmale, line 210: @pirupius did you just forget to update this file?

src/hooks/useFormComponent.tsx
```diff
@@ -0,0 +1,36 @@
...
+      };
+    });
+  }
+}
+
+const refs = {
+  pcr: 'component_preclinic-review',
+  hosp: 'component_hospitalization',
+  art: 'component_art',
+};
```
samuelmale, line 36: Oooppss! Is this object some kind of testing residue?

samuelmale (commented):
Thanks for working on this @pirupius; however, I have one concern. We seem to be duplicating efforts, especially in this specific task. 

[A backend module](https://github.com/openmrs/openmrs-module-o3forms) (currently bundled within the refapp) was introduced by @ibacher which defers most build tasks to the backend. This dramatically improves performance for big forms and I believe we should refactor the library to leverage this feature. This module offers the following:

- All referenced forms loaded into the appropriate parts of the main form
- All translations provided in a translation file for the appropriate locale
- Metadata about concepts used in the forms

@kajambiya - This module also plays a role tagential to i18n making life easy to go about adding support for internationalization.

--- 

@eudson @kajambiya @pirupius @ibacher  what do you think about adding support for "subforms" to this backend module and we get rid of the inbuilt form loader from the library?

kajambiya:
> Details



> Thanks for working on this @pirupius; however, I have one concern. We seem to be duplicating efforts, especially in this specific task.
> 
> [A backend module](https://github.com/openmrs/openmrs-module-o3forms) (currently bundled within the refapp) was introduced by @ibacher which defers most build tasks to the backend. This dramatically improves performance for big forms and I believe we should refactor the library to leverage this feature. This module offers the following:
> 
> * All referenced forms loaded into the appropriate parts of the main form
> * All translations provided in a translation file for the appropriate locale
> * Metadata about concepts used in the forms
> 
> @kajambiya - This module also plays a role tagential to i18n making life easy to go about adding support for internationalization.
> 
> @eudson @kajambiya @pirupius @ibacher what do you think about adding support for "subforms" to this backend module and we get rid of the inbuilt form loader from the library?

@samuelmale I think this is a nice module. Going to check it out especially in implementing form translations. The idea of using the module to load sub forms also sounds good especially as far as performance is concerned. Once we finish loading the components using the module, we shall look into expanding the model to add functionality for loading sub forms.

kajambiya:
@pirupius You have some lint and type checking issues.

ibacher:
> what do you think about adding support for "subforms" to this backend module and we get rid of the inbuilt form loader from the library?

@samuelmale @pirupius @eudson In principle, I'm onboard with bringing whatever functionality we can into the backend. However, I think there are some things we should consider here:

* What is the future goal we want; I think having both "references" and "sub-forms" is confusing. When, for example, would I use a "subform" instead of a reference to a page and vice versa?
* What is the "special functionality" that a "subform" might have?
* How are "subform" pages ordered with respect to the other pages?

Last point: I'd probably be happier with this if we can accomodate the "subform" terminology to something like the "reference" system. I realise that's a breaking change for OHRI forms, but that way, we avoid having two user-facing features were are a bit hard to disentangle. Ideally, we end up with a unified, intuitive single feature for including content from other forms in a single form and then for backwards compatibility, we can support "subform" as a synonym for some kind of reference with similar semantics.

eudson:
> > what do you think about adding support for "subforms" to this backend module and we get rid of the inbuilt form loader from the library?
> 
> @samuelmale @pirupius @eudson In principle, I'm onboard with bringing whatever functionality we can into the backend. However, I think there are some things we should consider here:
> 
> * What is the future goal we want; I think having both "references" and "sub-forms" is confusing. When, for example, would I use a "subform" instead of a reference to a page and vice versa? 
> * What is the "special functionality" that a "subform" might have?
> * How are "subform" pages ordered with respect to the other pages? 
> 
> Last point: I'd probably be happier with this if we can accomodate the "subform" terminology to something like the "reference" system. I realise that's a breaking change for OHRI forms, but that way, we avoid having two user-facing features were are a bit hard to disentangle. Ideally, we end up with a unified, intuitive single feature for including content from other forms in a single form and then for backwards compatibility, we can support "subform" as a synonym for some kind of reference with similar semantics.

@ibacher I think we will have to go with the suggestion of your last sentence. Sub-form it is not just a reusable component as the reference componentes are. When referencing sub-forms one can post data into multiple (different) `encounterTypes`. An example in OHRI is the COVID Assessment form, it has an embedded form for lab orders (not using the `order model` yet) that is used to do a lab order if certain conditions on the form are met. Another example is within HTS where if the client is HIV positive you want to do the enrolment right away, with sub-forms (I would call it embedded forms) you can do this seamlessly and  have the data posted into two different encounters: HIV Testing, ART Enrolment. This is just to mention a few examples. 

The Uganda team also wants to leverage this functionality. I believe this could have been also achieved with workflows but that feature is still not available in O3. cc: @slubwama 

Another cool thing of sub-forms is that they can also post to the same `encounterType` if you pass the right configs, right @samuelmale @kajambiya ? Which can essentially be considered a referenced component and could be expanded to become one, the only thing would be adding the cool stuff the AFE has where you can choose which parts to display and where.

Regarding supporting all of this from the backend and have all cooked when loading into the frontend I think it is a good idea and worths exploring. 

Can I propose we have a technical meeting about the whole form-engine architecture sooner than later because I keep seeing this comments on PRs and I think it would be beneficial if we agree on certain aspects before the code is written, @pirupius has spent a lot of time on this trying to figure out the best approach

.github/workflows/ci.yml
pirupius: This keeps failing cause the user (Lars) isn't part of the openmrs organization

src/hooks/useFormComponent.tsx
pirupius: Deleted

src/utils/forms-loader.ts
```diff
@@ -47,7 +47,7 @@ export function getForm(
 
 export function loadSubforms(parentForm) {
   parentForm.pages = parentForm.pages || [];
-  parentForm.pages.forEach(page => {
+  parentForm.pages.forEach((page) => {
```
pirupius, line 50: Same applies to this file. Only format changes

src/hooks/useFormJson.tsx
```diff
@@ -130,3 +136,76 @@ function setEncounterType(formJson: OHRIFormSchema): void {
     delete formJson.encounter;
   }
 }
+
+/**
+ * Functions to support reusable Form Components
+ */
+function getReferencedForms(formJson: OHRIFormSchema): Array<ReferencedForm> {
+  const referencedForms: Array<any> = formJson.referencedForms;
```
eudson, line 144: what happens when `formJson` is `undefined` ?

src/hooks/useFormJson.tsx
```diff
@@ -130,3 +136,76 @@ function setEncounterType(formJson: OHRIFormSchema): void {
...
+  if (!referencedForms) {
+    return [];
+  }
+  return referencedForms;
+}
+
+async function loadFormComponents(formComponentRefs: Array<ReferencedForm>): Promise<OHRIFormSchema[]> {
+  return Promise.all(formComponentRefs.map((formComponent) => loadFormJson(formComponent.formName, null, null)));
```
eudson, line 152: what happens when `formComponentRefs` is `undefined` ?

eudson (changes requested):

src/hooks/useFormJson.tsx
pirupius: This has been handled with the typings and validation on the call

pirupius:
we've created a ticket to explore migration to o3forms module as well as spec'ng migration of subform to references

pirupius:
@samuelmale @kajambiya please take another pass at this. If there no objections, will proceed and merge it and build on top of it for the other enhancements.

kajambiya (approved):
LGTM

pirupius:
Thanks @kajambiya 
Will go ahead and merge this. Post-merge reviews are welcome.

PR #153, opened 2023-12-12
(feat) Introduce generic program enrollment

kajambiya:
## Requirements

## Summary
This PR introduces generic program enrolment to the form engine. If an encounter is supposed to trigger a program enrolment, the user shall specify the same in the schema using `ProgramEnrollmentSubmissionAction` and specifying the program details through the config parameter. The user shall also use the enabled flag to specify whether the post submission is enabled or not. See sample below:
  ```
"postSubmissionActions": [
    {
      "actionId": "ProgramEnrollmentSubmissionAction",
      "enabled":"tbProgramType === '160541AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'",
      "config": {
        "enrollmentDate": "tbRegDate",
        "programUuid": "58005eb2-4560-4ada-b7bb-67a5cffa0a27",
        "completionDate": "outcomeTBRx"
      }
    }
]
```
## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/12844964/073c4286-140e-45b2-afcc-5fe18e743ee4

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -2,7 +2,7 @@ import { openmrsFetch, openmrsObservableFetch } from '@openmrs/esm-framework';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { encounterRepresentation } from '../constants';
-import { OpenmrsForm } from './types';
+import { OpenmrsForm, ProgramEnrollmentPayload } from './types';
 import { isUuid } from '../utils/boolean-utils';
 
 const BASE_WS_API_URL = '/ws/rest/v1/';
@@ -167,9 +167,9 @@ export function getPatientEnrolledPrograms(patientUuid: string) {
   });
 }
 
-export function createProgramEnrollment(payload, abortController) {
+export function createProgramEnrollment(payload: ProgramEnrollmentPayload, abortController: AbortController) {
   if (!payload) {
-    return null;
+    throw new Error('Program enrollment cannot be created because no payload is supplied');
   }
   const { program, patient, dateEnrolled, dateCompleted, location } = payload;
   return openmrsObservableFetch(`${BASE_WS_API_URL}programenrollment`, {
@@ -182,11 +182,15 @@ export function createProgramEnrollment(payload, abortController) {
   });
 }
 
-export function updateProgramEnrollment(programEnrollmentUuid: string, payload, abortController) {
-  if (!payload && !payload.program) {
-    return null;
+export function updateProgramEnrollment(
+  programEnrollmentUuid: string,
+  payload: ProgramEnrollmentPayload,
+  abortController: AbortController,
+) {
+  if (!payload || !programEnrollmentUuid) {
+    throw new Error('Program enrollment cannot be edited without a payload or a program Uuid');
   }
-  const { program, dateEnrolled, dateCompleted, location } = payload;
+  const { dateEnrolled, dateCompleted, location } = payload;
   return openmrsObservableFetch(`${BASE_WS_API_URL}programenrollment/${programEnrollmentUuid}`, {
     method: 'POST',
     headers: {
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -5,6 +5,8 @@ import { encounterRepresentation } from '../constants';
 import { OpenmrsForm } from './types';
 import { isUuid } from '../utils/boolean-utils';
 
+const BASE_WS_API_URL = '/ws/rest/v1/';
+
 export function saveEncounter(abortController: AbortController, payload, encounterUuid?: string) {
   const url = !!encounterUuid ? `/ws/rest/v1/encounter/${encounterUuid}?v=full` : `/ws/rest/v1/encounter?v=full`;
 
@@ -151,4 +153,46 @@ function dataURItoFile(dataURI: string) {
 
   const blob = new Blob([buffer], { type: mimeString });
   return blob;
+}
+
+//Program Enrollment
+export function getPatientEnrolledPrograms(patientUuid: string) {
+  return openmrsFetch(
+    `${BASE_WS_API_URL}programenrollment?patient=${patientUuid}&v=custom:(uuid,display,program,dateEnrolled,dateCompleted,location:(uuid,display))`,
+  ).then(({ data }) => {
+    if (data) {
+      return data;
+    }
+    return null;
+  });
+}
+
+export function createProgramEnrollment(payload, abortController) {
+  if (!payload) {
+    return null;
+  }
+  const { program, patient, dateEnrolled, dateCompleted, location } = payload;
+  return openmrsObservableFetch(`${BASE_WS_API_URL}programenrollment`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: { program, patient, dateEnrolled, dateCompleted, location },
+    signal: abortController.signal,
+  });
+}
+
+export function updateProgramEnrollment(programEnrollmentUuid: string, payload, abortController) {
+  if (!payload && !payload.program) {
+    return null;
+  }
+  const { program, dateEnrolled, dateCompleted, location } = payload;
+  return openmrsObservableFetch(`${BASE_WS_API_URL}programenrollment/${programEnrollmentUuid}`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: { dateEnrolled, dateCompleted, location },
+    signal: abortController.signal,
+  });
 }
--- a/src/ohri-form.component.test.tsx
+++ b/src/ohri-form.component.test.tsx
@@ -39,7 +39,7 @@ import labour_and_delivery_test_form from '../__mocks__/forms/ohri-forms/labour_
 import sample_fields_form from '../__mocks__/forms/ohri-forms/sample_fields.json';
 import postSubmission_test_form from '../__mocks__/forms/ohri-forms/post-submission-test-form.json';
 import * as registry from '../src/registry/registry';
-import { isPostSubmissionEnabled } from './utils/program-enrolment-helper';
+import { evaluatePostSubmissionExpression } from './utils/post-submission-action-helper';
 import * as formContext from './ohri-form-context';
 import * as usePostSubmission from './hooks/usePostSubmissionAction';
 
@@ -263,9 +263,9 @@ describe('OHRI Forms:', () => {
 
       const expression1 = "tbProgramType === '160541AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'";
       const expression2 = "tbProgramType === '160052AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'";
-      let enabled = isPostSubmissionEnabled(expression1, encounters);
+      let enabled = evaluatePostSubmissionExpression(expression1, encounters);
       expect(enabled).toEqual(true);
-      enabled = isPostSubmissionEnabled(expression2, encounters);
+      enabled = evaluatePostSubmissionExpression(expression2, encounters);
       expect(enabled).toEqual(false);
     });
     it('Should test post submission actions', async () => {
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: src/api/api.ts [hunk 1]
```diff
@@ -152,3 +154,45 @@ function dataURItoFile(dataURI: string) {
...
+    if (data) {
+      return data;
+    }
+    return null;
+  });
+}
+
+export function createProgramEnrollment(payload, abortController) {
+  if (!payload) {
+    return null;
```
ebambo, line 172: @kajambiya I think here you should throw an exception instead of returning null... you should return null when an object is not found (although you can also throw a NFE). Am saying this because the consumer of this method is expecting the created `programEnrolment` but since in this case the object will be null it will look like the enrolment was not created because God knows why but the real cause here is that he passed a null or undefined payload

src/api/api.ts
```diff
@@ -152,3 +154,45 @@ function dataURItoFile(dataURI: string) {
...
+  }
+  const { program, patient, dateEnrolled, dateCompleted, location } = payload;
```
ebambo, line 174: Here you are also assuming that all the properties of the payload exist, can't you validate this against a type ? something like: `payload instanceof ProgramEnrolmentType` ?

src/ohri-form.component.tsx
```diff
@@ -28,6 +28,7 @@ import LoadingIcon from './components/loaders/loading.component';
 import OHRIFormSidebar from './components/sidebar/ohri-form-sidebar.component';
 import WarningModal from './components/warning-modal.component';
 import styles from './ohri-form.component.scss';
+import { isPostSubmissionEnabled } from './utils/program-enrolment-helper';
```
ebambo, line 31: Why the `isPostSubmissionEnabled` helper part of the Program Enrolment Helpers? My recommendation is to put this helper on a more generic place  / file.

src/post-submission-actions/program-enrollment-action.ts
```diff
@@ -0,0 +1,97 @@
...
+                kind: 'error',
+                critical: false,
+                description: err?.message,
+              });
+            },
+          );
+        }
+      }
+    } else {
+      throw new Error('There is no program to enroll to');
```
ebambo, line 92: Please improve your message, this may led users/implementers to think there is no programs in the database whereas the issue here is that they did not provide a `programUUID`, unless am wrong.

src/utils/program-enrolment-helper.ts
```diff
@@ -0,0 +1,71 @@
+export function isPostSubmissionEnabled(expression: string, encounters: any[]): boolean {
```
ebambo, line 1: this function is doing more than checking if a `postSubmissionAction` is enabled... please rename this and also check if there is no function in the form engine that already does this... how `hideWhenExpression` is processed ?

ebambo (changes requested):
@kajambiya thank you for this but please have a look at my comments. Happy to discuss them

src/post-submission-actions/program-enrollment-action.ts
```diff
@@ -0,0 +1,97 @@
...
+            showToast({
+              title: 'Error saving program enrollment',
+              kind: 'error',
+              critical: false,
+              description: err?.message,
+            });
+          },
+        );
+      } else {
+        const patientEnrolledPrograms = await getPatientEnrolledPrograms(patient.id);
```
CynthiaKamau, line 60: `if (hasActiveEnrollment) {
        throw new Error('Cannot enroll patient to program. Patient already has an active enrollment');
 }`
Will this section get executed because you are checking if someone is already enrolled ?

CynthiaKamau (commented):
Great job :rocket:

src/api/api.ts
```diff
@@ -152,3 +154,45 @@ function dataURItoFile(dataURI: string) {
...
+export function getPatientEnrolledPrograms(patientUuid: string) {
+  return openmrsFetch(
+    `${BASE_WS_API_URL}programenrollment?patient=${patientUuid}&v=custom:(uuid,display,program,dateEnrolled,dateCompleted,location:(uuid,display))`,
+  ).then(({ data }) => {
+    if (data) {
+      return data;
+    }
+    return null;
+  });
+}
```
pirupius, line 168: Let's refactor this to use SWR

src/api/api.ts
```diff
@@ -152,3 +154,45 @@ function dataURItoFile(dataURI: string) {
...
+
+export function createProgramEnrollment(payload, abortController) {
```
pirupius, line 170: Let's define an interface for payload and add type checking to it

src/api/api.ts
```diff
@@ -152,3 +154,45 @@ function dataURItoFile(dataURI: string) {
...
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: { program, patient, dateEnrolled, dateCompleted, location },
+    signal: abortController.signal,
+  });
+}
+
+export function updateProgramEnrollment(programEnrollmentUuid: string, payload, abortController) {
```
pirupius, line 185: The typings have to be added here as well

src/ohri-form.component.tsx
```diff
@@ -210,14 +211,17 @@ const OHRIForm: React.FC<OHRIFormProps> = ({
...
                     if (encounterData.length) {
-                      await postAction.applyAction(
-                        {
-                          patient,
-                          sessionMode,
-                          encounters: encounterData,
-                        },
-                        config,
-                      );
+                      const isActionEnabled = enabled ? isPostSubmissionEnabled(enabled, encounterData) : true;
```
pirupius, line 214: Was logic meant to take care of backwards compatibility? if am reading this correctly if translates to if `enabled (true)` then do an action otherwise return `true` . So in both cases it resolves to `true`. If the condition `enabled` only applies to programs, shouldn't we consider adding it within the that handler to skip this step?

The main approach here assuming it's to either enroll or dis-enroll

src/post-submission-actions/program-enrollment-action.ts
```diff
@@ -0,0 +1,97 @@
...
+        let patientTBEnrollment = null;
```
pirupius, line 61: This looks like something you missed during the clean up. I don't think we should be adding program/implementation specific logic or naming here

pirupius (changes requested):

src/ohri-form.component.tsx
kajambiya: Hi @pirupius, if you remember that call we had with @ebambo, the `post submission action` is by default enabled, however it may be disabled given some conditions. See example below:
 ```
 "postSubmissionActions": [
    {
      "actionId": "ProgramEnrollmentSubmissionAction",
      "enabled":"tbProgramType === '160541AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'",
      "config": {
      "enrollmentDate": "tbRegDate",
      "programUuid": "58005eb2-4560-4ada-b7bb-67a5cffa0a27",
      "completionDate": "outcomeTBRx"
    }
  },
  {
    "actionId": "ProgramEnrollmentSubmissionAction",
    "enabled":"tbProgramType === '160052AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'",
    "config": {
    "enrollmentDate": "tbRegDate",
    "programUuid": "00f37871-0578-4ebc-af1d-e4b3ce75310d",
    "completionDate": "outcomeTBRx"
  }
}
  ],
```

We have 2 `post submission actions` but only one can run depending on the program type selected. The other will return `false` for the `enabled flag`

File: src/post-submission-actions/program-enrollment-action.ts (hunk 11 above)
Comment (kajambiya) [reply src/post-submission-actions/program-enrollment-action.ts:61]: Sure. Nice catch

File: src/post-submission-actions/program-enrollment-action.ts [hunk 12]
```diff
@@ -0,0 +1,97 @@
...
+    const completionDate = encounter.obs?.find((item) => item.formFieldPath.includes(config.completionDate))?.value;
+    const programUuid = config.programUuid;
+
+    if (programUuid) {
+      const abortController = new AbortController();
+      const payload = {
+        patient: patient.id,
+        program: programUuid,
+        dateEnrolled: enrollmentDate ? dayjs(enrollmentDate).format() : null,
+        dateCompleted: completionDate ? dayjs(enrollmentDate).format() : null,
```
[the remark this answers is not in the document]
kajambiya: So the api works in such a way that when the completion date is supplied, the program status changes from` Active `to `Completed`. So to answer your question, a user is disenrolled from a program the moment a completion date is supplied.

src/api/api.ts
kajambiya: Yah, makes sense

src/post-submission-actions/program-enrollment-action.ts
kajambiya: Not sure I fully get the question but let me briefly explain what is happening here;
So there's an existing module through which patients are being enrolled into programs. However we are also adding provision for enrolling patients into programs after successful submission of some encounters using post submission actions. Now when the post submission action is in `enter mode`, the assumption is that someone is being enrolled into a program for the 1st time, however if they already has an active program of the same type, they can't be enrolled again. This is being handled by the backend but saw it wise to handle it on the frontend as well. And yah, the section will get executed.

src/api/api.ts
kajambiya: Nice one

src/utils/program-enrolment-helper.ts
kajambiya: There's an existing `expression-runner` but after studying it, I realised it couldn't be reused for this scenario:
The expression runner has access to all fields so expressions are evaluated in relation to all dependent fields regardless of whether they have data or not. The `post submission action` on the other hand has no access to all fields but has access to the submitted encounter and the submitted encounter eliminates fields/obs that don't have values, so it's just a few fields that get submitted.
The expression-runner also uses the `encounter context` which is not present at post-submission.

src/post-submission-actions/program-enrollment-action.ts
kajambiya: Sure

src/api/api.ts
kajambiya: I feel like SWR would be an overkill here since the dataset will always be small and we are not making many calls to the api.

src/ohri-form.component.tsx
kajambiya: Done

src/api/api.ts
kajambiya: Done

src/post-submission-actions/program-enrollment-action.ts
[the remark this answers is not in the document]
kajambiya: I think I get what you meant. I was passing `enrollmentDate` as the value of` dateCompleted`. This is now fixed.

src/utils/program-enrolment-helper.ts
ebambo: @kajambiya but the name of this function is not inline with what it does.

src/ohri-form.component.test.tsx
```diff
@@ -234,10 +241,98 @@ describe('OHRI Forms:', () => {
...
+                value: '2023-12-05T00:00:00.000+0000',
+              },
+            ],
+          },
+        ],
+      });
+
+      // Render the form
+      await act(async () => renderForm(null, postSubmission_test_form));
+      //const getRegisteredPostSubmissionActionSpy = jest.spyOn(registry, 'getRegisteredPostSubmissionAction');
```
pirupius, line 311: Do we need this? If not then let's clean this up

src/ohri-form.component.test.tsx
```diff
@@ -234,10 +241,98 @@ describe('OHRI Forms:', () => {
...
+      // Simulate a successful form submission
+      await act(async () => {
+        fireEvent.submit(screen.getByText(/save/i));
+      });
+
+      expect(saveEncounterMock).toHaveBeenCalled();
+      await act(async () => expect(saveEncounterMock).toReturn());
+      //expect(postSubmissionSpy).toHaveBeenCalled();
+
+      //expect(api.createProgramEnrollment).toHaveBeenCalled();
```
pirupius, line 330: same applies to this

pirupius (approved):

ebambo (approved):

PR #156, opened 2023-12-15
(feat) validation of obs group count

arodidev:
## Requirements

## Summary

This PR introduces validation that checks whether the number of repeating obsGroup fields matches the number count specified.
## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/51090527/da6959b4-4ca6-4df1-bd6f-cb2217462221

### DIFF
--- a/src/components/encounter/ohri-encounter-form.component.tsx
+++ b/src/components/encounter/ohri-encounter-form.component.tsx
@@ -91,7 +91,7 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
   const [isFieldInitializationComplete, setIsFieldInitializationComplete] = useState(false);
   const [invalidFields, setInvalidFields] = useState([]);
   const [initValues, setInitValues] = useState({});
-  const [obsGroupCounter, setObsGroupCounter] = useState<Array<RepeatObsGroupCounter | null>>([]);
+  const [obsGroupCounter, setObsGroupCounter] = useState<Array<RepeatObsGroupCounter>>([]);
   const layoutType = useLayoutType();
 
   const encounterContext = useMemo(
--- a/src/components/encounter/ohri-encounter-form.component.tsx
+++ b/src/components/encounter/ohri-encounter-form.component.tsx
@@ -1,5 +1,5 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import { SessionLocation, useLayoutType, Visit } from '@openmrs/esm-framework';
+import React, { Dispatch, SetStateAction, useCallback, useEffect, useMemo, useState } from 'react';
+import { SessionLocation, showToast, useLayoutType, Visit } from '@openmrs/esm-framework';
 import { ConceptFalse, ConceptTrue } from '../../constants';
 import {
   OHRIFormField,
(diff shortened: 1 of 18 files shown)


src/components/encounter/ohri-encounter-form.component.tsx
```diff
@@ -88,6 +91,7 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
   const [isFieldInitializationComplete, setIsFieldInitializationComplete] = useState(false);
   const [invalidFields, setInvalidFields] = useState([]);
   const [initValues, setInitValues] = useState({});
+  const [obsGroupCounter, setObsGroupCounter] = useState<Array<RepeatObsGroupCounter | null>>([]);
```
pirupius, line 94: The `null` here feels redundant. is there a specific reason why u added it?

src/ohri-form.component.test.tsx
```diff
@@ -232,6 +236,50 @@ describe('OHRI Forms:', () => {
     });
   });
 
+  describe('Labour and delivery validation', () => {
+    fit('should show error toast when the birth count does not match birth info count', async () => {
```
pirupius, line 240: Please rename these tests to something more generic. I believe the naming could be something about verifying obs group count

src/ohri-form.component.tsx
```diff
@@ -332,6 +332,7 @@ const OHRIForm: React.FC<OHRIFormProps> = ({
                         handlers={handlers}
                         workspaceLayout={workspaceLayout}
                         isSubmitting={isSubmitting}
+                        setIsSubmitting={setIsSubmitting}
```
pirupius, line 335: Am curious what the difference between the 2 will be. Don't they do similar things?

src/ohri-form-context.tsx
```diff
@@ -26,6 +26,7 @@ export interface EncounterContext {
   encounterDate: Date;
   setEncounterDate(value: Date): void;
   initValues?: Record<string, any>;
+  setObsGroupCounter?: any;
```
pirupius, line 29: What happens if there's more than one obs group field?

pirupius (changes requested):

src/components/encounter/ohri-encounter-form.component.tsx
arodidev: In the event that there are no obsGroups in the form, this array will always remain empty.

src/ohri-form-context.tsx
arodidev: Each will be handled individually, although I am not quite sure I understand what this question asks.

src/ohri-form.component.test.tsx
arodidev: Noted.

src/ohri-form.component.tsx
arodidev: I am not sure I understand what you are alluding to, kindly explan.

src/components/encounter/ohri-encounter-form.component.tsx
pirupius: I think the initialization u added already takes care of that so there's no need to the null

src/ohri-form-context.tsx
pirupius: You are passing `setObsGroupCounter` on the entire form whether there's need to track count or not. In a scenario where this is being referenced more than one, it might bring a problem especially because you are tracking it at the form context level

arodidev: Oh, I see what you mean now. So yes the `setObsGroupCounter` will only update the state with objects of type repeatObsGroupCounter, see [here](https://arc.net/l/quote/iczqbmqa), this object will always include the field id for identification purposes, and in the event the obsGroup is duplicated it will always push the object with the most recent count to the front of the array.

src/components/encounter/ohri-encounter-form.component.tsx
arodidev: Noted.

pirupius (approved):

PR #168, opened 2024-02-06
Form Engine UI refactor

arodidev:
## Requirements

## Summary

This PR aims to tackle various issues:
- Responsiveness of the controls rendered on the form.
- Refactoring of previous value out of the control component and made to render independently
- Fixing previous value for various fields, missing or non-functional
- Refactor controls to pull cross-cutting styles from a single stylesheet
- Remove previous value functionality from obsGroup fields.
- Refactoring questionInfo tooltip positions for controls
- Add functionality to dynamically set control width from the form JSON
- Fixing control misalignment within the obsGroups

## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/51090527/84df4ee9-4884-4fec-8e09-99417525913b


src/components/section/ohri-form-section.component.tsx
```diff
@@ -1,23 +1,32 @@
...
 interface FieldComponentMap {
   fieldComponent: React.ComponentType<OHRIFormFieldProps>;
   fieldDescriptor: OHRIFormField;
   handler: SubmissionHandler;
 }
 
+//move this to helper file
+function previousValueDisplayForCheckbox(previosValueItems: Object[]): String {
+  return previosValueItems.map((eachItem) => eachItem['display']).join(', ');
+}
```
pirupius, line 24: Is this being reused else where?

src/components/section/ohri-form-section.component.tsx
```diff
@@ -38,30 +47,85 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
           .filter((entry) => entry?.fieldComponent)
           .map((entry, index) => {
             const { fieldComponent: FieldComponent, fieldDescriptor, handler } = entry;
+
+            const prevValue = encounterContext.previousEncounter
```
pirupius, line 51:
```suggestion
            const previousValue = encounterContext.previousEncounter
```

src/components/section/ohri-form-section.component.tsx
```diff
@@ -38,30 +47,85 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
...
+              ? handler?.getPreviousValue(fieldDescriptor, encounterContext.previousEncounter, fieldsFromEncounter)
+              : { value: 'no previous value' };
+
             if (FieldComponent) {
+              const previousValue = previousValues?.find((searchItem) => searchItem.field === fieldDescriptor.id);
```
pirupius, line 56: This const `previousValue` naming is being duplicated

src/components/section/ohri-form-section.component.tsx
```diff
@@ -38,30 +47,85 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
...
                   question={fieldDescriptor}
                   onChange={onFieldChange}
                   key={index}
                   handler={handler}
                   useField={useField}
+                  previousValue={previousValue}
                 />
               );
 
+              const prevCheckboxDisplayValue = Array.isArray(prevValue)
```
pirupius, line 68:
```suggestion
              const previousCheckboxDisplayValue = Array.isArray(previousValue)
```

src/components/section/ohri-form-section.component.tsx
```diff
@@ -38,30 +47,85 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
...
+                ? previousValueDisplayForCheckbox(prevValue)
```
pirupius, line 69:
```suggestion
                ? previousValueDisplayForCheckbox(previousValue)
```

src/components/section/ohri-form-section.component.tsx
```diff
@@ -38,30 +47,85 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
...
+                    prevValue &&
+                    !isTrue(fieldDescriptor.questionOptions.usePreviousValueDisabled) && (
+                      <div className={styles.previousValue}>
+                        <PreviousValueReview
+                          value={prevValue}
+                          displayText={
+                            fieldDescriptor.questionOptions.rendering == 'checkbox'
+                              ? prevCheckboxDisplayValue
+                              : prevValue?.display
+                          }
```
pirupius, line 123: Pay attention to the usages of `previousValue` and `prevValue` when aligning the naming.

src/components/section/ohri-form-section.scss
```diff
@@ -1,49 +1,77 @@
 @use '@carbon/colors';
 
-.section {
-  border-radius: solid 1px rgba(107, 104, 104, 0.5);
-  margin-bottom: 3.063rem;
+.parentResizer{
+margin-top: 0.5rem;
+margin-bottom: 1.5rem;
```
pirupius, line 5: This indentation looks off

src/ohri-form.component.scss
```diff
@@ -24,7 +24,7 @@
 
 .ohriFormContainer :global(.cds--label) {
   font-size: 0.875rem;
-  width: 18rem;
+  // width: 18rem;
```
pirupius, line 27: more residue to delete

pirupius (commented):
Nice work with the clean up and refactors. I have some concerns about the naming and duplication of variable names

src/components/inputs/text/ohri-text.component.tsx
```diff
@@ -73,18 +70,18 @@ const OHRIText: React.FC<OHRIFormFieldProps> = ({ question, onChange, handler })
...
-            }>
+            }`}>
             <TextInput
               {...field}
               id={question.id}
               labelText={question.label}
               name={question.id}
               value={field.value || ''}
-              onFocus={() => setPreviousValue(field.value)}
+              onFocus={''}
```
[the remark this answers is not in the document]
arodidev: Based on how the current implementation of previous value is on the section, this method of updating previous value using control level state is now deprecated

src/components/inputs/ui-select-extended/ui-select-extended.scss
```diff
@@ -1,55 +1,55 @@
...
+// .formField > div > div > label {
+//   color: colors.$gray-70;
+//   }
   
-.multiselectOverride {
-  width: auto;
-  padding-bottom: 15px;
-  }
+// .multiselectOverride {
+//   width: auto;
```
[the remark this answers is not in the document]
arodidev: Nice catch, still have some cleaning up to do, thanks.

src/components/section/ohri-form-section.scss
```diff
@@ -1,49 +1,77 @@
 @use '@carbon/colors';
 
-.section {
-  border-radius: solid 1px rgba(107, 104, 104, 0.5);
-  margin-bottom: 3.063rem;
+.parentResizer{
```
samuelmale, line 3: Is this a cross-cutting styling or it's an obs group style? Plus for more consistency, let's properly format SCSS files:

```css
.myClassName {
```
👍 

```css
.myClassName{
```
👎

src/submission-handlers/base-handlers.ts
```diff
@@ -6,6 +6,7 @@ import { OHRIFormField, OpenmrsEncounter, OpenmrsObs, SubmissionHandler } from '
 import { parseToLocalDateTime } from '../utils/ohri-form-helper';
 import { flattenObsList, hasRendering } from '../utils/common-utils';
 import { formatDate } from '@openmrs/esm-framework';
+import { render } from '@testing-library/react';
```
samuelmale, line 9: Do you need this?

src/utils/common-expression-helpers.ts
```diff
@@ -468,3 +469,17 @@ export function registerDependency(node: FormNode, determinant: OHRIFormField) {
...
+  if (!booleanConceptRepresentation) {
+    throw new Error('booleanConceptRepresentation cannot be a null value');
+  }
+  if (booleanConceptRepresentation == ConceptTrue) {
+    return true;
+  }
+  if (booleanConceptRepresentation == ConceptFalse) {
+    return false;
+  }
+};
```
samuelmale, line 485: There are two issues here:
1. This seems to be out of scope and ideally should be addressed in a follow-up PR
2. This seems like an Implementation detail and if we have to support this, it should be driven through a configuration.

samuelmale (changes requested):
Nice work @arodidev, just a few comments. Maybe as a general comment: There is no linked ticket to define the scope of work which makes the review process tricky. Plus I'm in agreement with harmonizing the cross-cutting stylings but styles specific to input components shouldn't be migrated to the `ohri-form-section.sccs` file.

src/api/types.ts
```diff
@@ -131,6 +132,7 @@ export interface OHRIFormFieldProps {
   handler: SubmissionHandler;
   // This is of util to components defined out of the engine
   useField?: (fieldId: string) => [FieldInputProps<any>, FieldMetaProps<any>, FieldHelperProps<any>];
+  previousValue?: Record<string, any>;
```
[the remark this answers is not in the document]
arodidev: I agree, maybe also including an additional type since `multi-select` receives previous value as an array of values rather than an individual value.

src/components/group/ohri-obs-group.scss
```diff
@@ -1,48 +0,0 @@
...
-  display: flex;
-  justify-content: space-between;
-  align-items: center;
-}
-
-.tooltip{
-  display: flex;
-  justify-content: flex-end;
-  align-items: center;
-  margin-top: 0.5rem;
```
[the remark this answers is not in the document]
arodidev: So the idea of this was to undo all styling from the controls and to start the styling process from scratch centrally, as there were a lot of deprecated/irrelevant styling within the many stylesheets. I agree that the cleanup could have been handled better.

src/components/section/ohri-form-section.scss
```diff
@@ -1,49 +1,77 @@
...
 }
+
+.datetime{
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+
+  > .timePickerSpacing{
+    margin-left: 1rem;
+  }
```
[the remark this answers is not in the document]
arodidev: I agree there will be some degree of independence regarding the individual input component styles, what do you suggest as a solution to tackle this? Reinstating the individual stylesheets?

src/utils/common-expression-helpers.ts
arodidev: Out of curiosity, what is the rationale for this? Is it that there could be multiple concepts for true/ false?

samuelmale: My point is if we are to support multiple implementations, we should not hardcode refs to non-system level metadata. In short, the `ConceptTrue` vs `ConceptFalse` may vary across downstream distros.

src/components/group/ohri-obs-group.scss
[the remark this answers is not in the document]
samuelmale: While I'm on board with the clean-up initiative, I'd like to emphasize that it's generally considered good practice to maintain specific CSS rules within component-specific stylesheets rather than consolidating everything into one large CSS file. This approach helps to keep things more granular and makes maintenance easier in the long run.

src/components/section/ohri-form-section.scss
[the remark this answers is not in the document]
samuelmale: Let's stick to migrating cross-cutting styles

src/utils/common-expression-helpers.ts
samuelmale: Maybe it's a good idea to register this helper via the registry?

src/components/section/ohri-form-section.component.tsx
arodidev: No, just the one instance

src/utils/common-expression-helpers.ts
eudson: @samuelmale I agree this is out of the scope of this PR. But I thought the conceptTrue / conceptFalse issue was a common problem... and also I don't see any conceptUUIDs here @arodidev how does this comparison really works?

arodidev: @eudson the function compares the argument passed to it with conceptTrue and conceptFalse which are hard-coded concept UUIDs for true/ false.

src/components/section/ohri-form-section.component.tsx
```diff
@@ -38,30 +47,85 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
...
             if (FieldComponent) {
+              const previousValue = previousValues?.find((searchItem) => searchItem.field === fieldDescriptor.id);
               const qnFragment = (
                 <FieldComponent
                   question={fieldDescriptor}
                   onChange={onFieldChange}
                   key={index}
                   handler={handler}
                   useField={useField}
+                  previousValue={previousValue}
```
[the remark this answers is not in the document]
arodidev: I have a few questions regarding this approach...

- If we are picking the field's previous value from the previous encounter, what is the reason for subjecting the values to form validators again if these were already submitted, presumably having passed the validation before being submitted?
- What are the improvements you suggest in order to better this approach?
- Once this implementation is in, while migrating to React Form Hook, are we simply going to scrub it in favour of a new implementation? Because in that case I feel it wiser to wait until the migration in order to handle this any which way, no?

src/api/types.ts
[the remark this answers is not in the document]
samuelmale: @arodidev why not simply:

```js
previousValue: string | number | Date | boolean | [] | undefined;
```

Or even better:

```js
previousValue: FormFieldValue;
```

src/components/section/ohri-form-section.component.tsx
```diff
@@ -1,23 +1,32 @@
...
   handler: SubmissionHandler;
 }
 
+//move this to helper file
+function previousValueDisplayForCheckbox(previosValueItems: Object[]): String {
+  return previosValueItems.map((eachItem) => eachItem['display']).join(', ');
+}
+
 const OHRIFormSection = ({ fields, onFieldChange }) => {
+  const [previousValues, setPreviousValues] = useState<Array<Record<string, any>>>([]);
```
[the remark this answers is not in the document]
samuelmale: It's cheaper to index values from a map as opposed to searching an array: 

```js

previousValuesMap[fieldDescriptor.id]
```
vs

```js

previousValuesArray.find(candidate => candidate.field === fieldDescriptor.id)
```

src/components/section/ohri-form-section.component.tsx
```diff
@@ -1,23 +1,32 @@
 import React, { useEffect, useState } from 'react';
 import { useField } from 'formik';
 import { ErrorBoundary } from 'react-error-boundary';
-import { ToastNotification } from '@carbon/react';
+import { ToastNotification, Grid, Column, Row } from '@carbon/react';
```
samuelmale, line 4: Do you need the change above?

src/components/section/ohri-form-section.component.tsx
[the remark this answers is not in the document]
samuelmale: > If we are picking the field's previous value from the previous encounter, what is the reason for subjecting the values to form validators again if these were already submitted, presumably having passed the validation before being submitted?

- **Existence of time-sensitive business logic**: Certain data may have expiration dates or validity periods, and revalidation ensures that the information remains accurate and up-to-date.
- **Revised business logic for new form iterations**: In cases where we have new form versions, the validation criteria or rules may also change. Revalidating previously submitted values ensures that they conform to the latest business logic and validation rules.
- **Revalidation considering changes in related values**: Business logic may depend on the relationships between different data points or values. If other related values have changed since the previous encounter, it may affect the validity or interpretation of the previously submitted values.

> Because in that case I feel it wiser to wait until the migration in order to handle this any which way, no?

I would say yes and no! Yes, because we ought to migrate anyway. No, because I believe there should be a simpler way of achieving the UI fixes without this expensive approach. Consider the transition to Form Hook as a potential breaking change that will require a major release.

Hear me out: As much as this doesn't seem like a breaking change, I would say it's a major change that may introduce unknown side effects. This doesn't seem like a major revamp that addresses performance-related issues or restructuring components to make maintenance simpler, but rather a UX bump.

Question: Is it possible to address the UI issues without resorting to this approach? *(I'm not conservative about the legacy patterns but rather a little worried about how we are approaching things here.)*

While it makes sense for the section (parent) to manage the rendering of its questions (children), I don't think it's a good idea for it to directly manage the states of their previous values. In my pov, I would suggest addressing the UI issues tangential to previous values by using robust CSS selectors instead of the DOM shift. We can then set up an epic to address most of these technical debts, such as decoupling validation and business logic processing from the [ohri-form.component.tsx](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/ohri-form.component.tsx).

src/components/section/ohri-form-section.component.tsx
arodidev: Will resolve

src/components/section/ohri-form-section.component.tsx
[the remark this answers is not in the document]
arodidev: While I do agree that this is a large change, I still don't believe there is a feasible way to achieve this with only CSS, unless of course you are open to discourse and finding a solution we can both agree on?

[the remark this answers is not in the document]
eudson: @samuelmale as much as we appreciate all the valuable comments/inputs given to date. We need this work merged for because we need it for our next release. Regarding the approaches you are proposing I would suggest you propose them in form of new tickets on the [React Form Engine NEXT](https://openmrs.atlassian.net/browse/O3-2193)  epic on Jira. Unless there is an obvious bug lets merge this PR please. Am also open to have a call to discuss this. Thank you

eudson (approved):
Lets merge this and have the enhancements ticketed for a different PR.

CynthiaKamau:
Thanks @arodidev , kindly fix the conflicts

CynthiaKamau (approved):

src/components/inputs/multi-select/ohri-multi-select.component.tsx
```diff
@@ -114,9 +119,9 @@ export const OHRIMultiSelect: React.FC<OHRIFormFieldProps> = ({ question, onChan
             readOnly={question.readonly}
           />
         </div>
-        <div className={styles.formField} style={{ marginTop: '0.125rem' }}>
+        <div style={{ marginTop: '0.125rem' }}>
```
pirupius, line 122: Is this a style that is used in other controls? Let's extract this to the scss file

src/components/inputs/radio/ohri-radio.scss
```diff
@@ -1,35 +1,27 @@
...
-
-.formField > div > div > label {
-  color: colors.$gray-70;
-}
-
-.row {
-  display: flex;
-  flex-direction: row;
-  align-items: center;
+.errorLegend legend {
```
pirupius, line 3: Indentation on this file looks off. Please format this to 2 spaces

src/components/inputs/select/ohri-dropdown.scss
```diff
@@ -1,35 +1,11 @@
...
-
-  > div > div > div {
-    width: 22.313rem !important;
-  }
-}
-.row {
-  display: flex;
-  flex-direction: row;
-  align-items: baseline;
+.boldedLabel label {
```
pirupius, line 3: same applies to this file around indenting

src/components/inputs/ui-select-extended/ui-select-extended.tsx
```diff
@@ -69,6 +69,16 @@ const UISelectExtended: React.FC<OHRIFormFieldProps> = ({ question, handler, onC
     question.value = handler?.handleFieldSubmission(question, value, encounterContext);
   };
 
+  useEffect(() => {
+    if (!isEmpty(previousValue)) {
```
pirupius, line 73: Is this a control we want to add a previous value to? So much is happening already with the search and might have some points of failure or conflict

pirupius (commented):
Some final touch up comments

src/api/types.ts
```diff
@@ -117,8 +117,12 @@ export interface OHRIFormField {
   validators?: Array<Record<string, any>>;
   behaviours?: Array<Record<string, any>>;
   questionInfo?: string;
+  constrainMaxWidth?: boolean;
+}
+export interface previousValue {
+  field: string;
+  value: string | number | Date | boolean | previousValue[];
```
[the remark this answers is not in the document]
arodidev: The `multi-select` input component receives its values as an array of type `previousValue`

src/components/inputs/content-switcher/ohri-content-switcher.component.tsx
```diff
@@ -20,6 +21,15 @@ export const OHRIContentSwitcher: React.FC<OHRIFormFieldProps> = ({ question, on
     }
   }, [question]);
 
+  useEffect(() => {
+    if (!isEmpty(previousValue)) {
+      const { value } = previousValue;
```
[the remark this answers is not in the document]
arodidev: Due to the differences in datatypes of the `previousValues` for specific controls, otherwise I think the format is pretty standard across the board, maybe expound a bit?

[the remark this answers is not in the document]
pirupius: This isn't a show stopper. There other controls that access it with `previousValue.value` and others do is this way based on your PR

src/components/inputs/ui-select-extended/ui-select-extended.tsx
arodidev: I see, although `previousValue` was already supported by `ui-select-extended` prior to this PR.

PR #174, opened 2024-03-12
(feat) Introduce Encounter Provider Control

kajambiya:
## Requirements

## Summary
This PR introduces the Encounter Provider control.
When submitting an encounter, RFE by default picks the provider from the session(logged in user) however there are scenario where the provider is passed as part of the encounter. In this case, the default provider needs to be overridden. 

## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/12844964/c48e5c54-1ed9-40c7-84f7-1afeb0ac3409

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -71,6 +71,12 @@ export function getAllLocations(): Observable<{ uuid: string; display: string }[
   );
 }
 
+export function getEncounterProviders(): Observable<{ uuid: string; display: string }[]> {
+  return openmrsObservableFetch(`/ws/rest/v1/provider?v=custom:(uuid,display)`).pipe(
+    map(({ data }) => data['results']),
+  );
+}
+
 export async function getPreviousEncounter(patientUuid: string, encounterType: string) {
   const query = `patient=${patientUuid}&_sort=-_lastUpdated&_count=1&type=${encounterType}`;
   let response = await openmrsFetch(`/ws/fhir2/R4/Encounter?${query}`);
--- a/src/components/inputs/ui-select-extended/ui-select-extended.tsx
+++ b/src/components/inputs/ui-select-extended/ui-select-extended.tsx
@@ -45,7 +45,10 @@ const UISelectExtended: React.FC<OHRIFormFieldProps> = ({ question, handler, onC
   }, [encounterContext.sessionMode, question.readonly, question.inlineRendering, layoutType, workspaceLayout]);
 
   useEffect(() => {
-    const datasourceName = question.questionOptions?.datasource?.name;
+    let datasourceName = question.questionOptions?.datasource?.name;
+    if (question.type === 'encounterProvider') {
+      datasourceName = question.type;
+    }
     setConfig(
       datasourceName
         ? question.questionOptions.datasource?.config
--- a/src/components/section/helpers.ts
+++ b/src/components/section/helpers.ts
@@ -21,7 +21,9 @@ export function getFieldControlWithFallback(question: OHRIFormField) {
   if (question.type === 'encounterLocation') {
     question.questionOptions.rendering = 'encounter-location';
   }
-
+  if (question.type === 'encounterProvider') {
+    return getRegisteredControl('encounterProvider');
+  }
   // Retrieve the registered control based on the specified rendering
   return getRegisteredControl(question.questionOptions.rendering);
 }
--- a/src/datasources/provider-datasource.ts
+++ b/src/datasources/provider-datasource.ts
@@ -1,9 +1,9 @@
 import { openmrsFetch } from '@openmrs/esm-framework';
 import { BaseOpenMRSDataSource } from './data-source';
 
-export class LocationDataSource extends BaseOpenMRSDataSource {
+export class ProviderDataSource extends BaseOpenMRSDataSource {
   constructor() {
-    super('/ws/rest/v1/location?v=custom:(uuid,display)');
+    super('/ws/rest/v1/provider?v=custom:(uuid,display)');
   }
 
   fetchData(searchTerm: string, config?: Record<string, any>, uuid?: string): Promise<any[]> {
@@ -19,6 +19,7 @@ export class LocationDataSource extends BaseOpenMRSDataSource {
 
     return openmrsFetch(searchTerm ? `${apiUrl}&q=${searchTerm}` : apiUrl).then(({ data }) => {
       if (data.results) {
+        console.log(data.results);
         return data.results;
       }
       return data;
--- a/src/submission-handlers/encounterProviderHandler.ts
+++ b/src/submission-handlers/encounterProviderHandler.ts
@@ -1,0 +1,24 @@
+import { SubmissionHandler } from '..';
+import { getEncounterProviders } from '../api/api';
+import { OpenmrsEncounter, OHRIFormField } from '../api/types';
+import { EncounterContext } from '../ohri-form-context';
+
+export const EncounterProviderHandler: SubmissionHandler = {
+  handleFieldSubmission: (field: OHRIFormField, value: any, context: EncounterContext) => {
+    context.setEncounterProvider(value);
+    return value;
+  },
+  getInitialValue: (encounter: OpenmrsEncounter, field: OHRIFormField, allFormFields?: OHRIFormField[]) => {
+    return new Date(); // TO DO: pick it from the visit if present
+  },
+
+  getDisplayValue: (field: OHRIFormField, value: any) => {
+    if (!field.value) {
+      return null;
+    }
+    return value;
+  },
+  getPreviousValue: (field: OHRIFormField, encounter: OpenmrsEncounter, allFormFields: Array<OHRIFormField>) => {
+    return null;
+  },
+};
--- a/src/components/encounter/ohri-encounter-form.component.tsx
+++ b/src/components/encounter/ohri-encounter-form.component.tsx
@@ -83,6 +83,7 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
   const [fields, setFields] = useState<Array<OHRIFormField>>([]);
   const [encounterLocation, setEncounterLocation] = useState(null);
   const [encounterDate, setEncounterDate] = useState(formSessionDate);
+  const [encounterProvider, setEncounterProvider] = useState(provider);
   const { encounter, isLoading: isLoadingEncounter } = useEncounter(formJson);
   const [previousEncounter, setPreviousEncounter] = useState<OpenmrsEncounter>(null);
   const [isLoadingPreviousEncounter, setIsLoadingPreviousEncounter] = useState(true);
@@ -92,6 +93,7 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
   const [invalidFields, setInvalidFields] = useState([]);
   const [initValues, setInitValues] = useState({});
   const [obsGroupCounter, setObsGroupCounter] = useState<Array<RepeatObsGroupCounter>>([]);
+
   const layoutType = useLayoutType();
 
   const encounterContext = useMemo(
@@ -102,9 +104,11 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
       location: location,
       sessionMode: sessionMode || (form?.encounter ? 'edit' : 'enter'),
       encounterDate: formSessionDate,
+      encounterProvider: provider,
       form: form,
       visit: visit,
       setEncounterDate,
+      setEncounterProvider,
       initValues: initValues,
       obsGroupCounter: obsGroupCounter,
       setObsGroupCounter: setObsGroupCounter,
@@ -478,13 +482,13 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
       // update encounter providers
       const hasCurrentProvider =
         encounterForSubmission['encounterProviders'].findIndex(
-          (encProvider) => encProvider.provider.uuid == provider,
+          (encProvider) => encProvider.provider.uuid == encounterProvider,
         ) !== -1;
       if (!hasCurrentProvider) {
         encounterForSubmission['encounterProviders'] = [
           ...encounterForSubmission.encounterProviders,
           {
-            provider: provider,
+            provider: encounterProvider,
             encounterRole: encounterRole?.uuid,
           },
         ];
@@ -504,7 +508,7 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
         encounterType: formJson.encounterType,
         encounterProviders: [
           {
-            provider: provider,
+            provider: encounterProvider,
             encounterRole: encounterRole?.uuid,
           },
         ],
@@ -709,6 +713,7 @@ export const OHRIEncounterForm: React.FC<OHRIEncounterFormProps> = ({
       value={{
         values,
         setFieldValue,
+        setEncounterProvider: setEncounterProvider,
         setEncounterLocation: setEncounterLocation,
         setObsGroupsToVoid: setObsGroupsToVoid,
         obsGroupsToVoid: obsGroupsToVoid,
--- a/src/datasources/concept-data-source.ts
+++ b/src/datasources/concept-data-source.ts
@@ -15,7 +15,7 @@ export class ConceptDataSource extends BaseOpenMRSDataSource {
       } else {
         return openmrsFetch(searchTerm ? `${apiUrl}&q=${searchTerm}` : apiUrl).then(({ data }) => {
           return data.results.filter(
-            concept => concept.conceptClass && config.class.includes(concept.conceptClass.uuid),
+            (concept) => concept.conceptClass && config.class.includes(concept.conceptClass.uuid),
           );
         });
       }
--- a/src/hooks/useEncounterRole.tsx
+++ b/src/hooks/useEncounterRole.tsx
@@ -6,7 +6,7 @@ export function useEncounterRole() {
     '/ws/rest/v1/encounterrole?v=custom:(uuid,display,name)',
     openmrsFetch,
   );
-  const clinicalEncounterRole = data?.data.results.find(encounterRole => encounterRole.name === 'Clinician');
+  const clinicalEncounterRole = data?.data.results.find((encounterRole) => encounterRole.name === 'Clinician');
 
   if (clinicalEncounterRole) {
     return { encounterRole: clinicalEncounterRole, error, isLoading };
--- a/src/ohri-form-context.tsx
+++ b/src/ohri-form-context.tsx
@@ -6,6 +6,7 @@ type OHRIFormContextProps = {
   values: Record<string, any>;
   setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void;
   setEncounterLocation: (value: any) => void;
+  setEncounterProvider: (value: any) => void;
   obsGroupsToVoid: Array<any>;
   setObsGroupsToVoid: (value: any) => void;
   encounterContext: EncounterContext;
@@ -25,6 +26,8 @@ export interface EncounterContext {
   sessionMode: SessionMode;
   encounterDate: Date;
   setEncounterDate(value: Date): void;
+  encounterProvider: string;
+  setEncounterProvider(value: string): void;
   initValues?: Record<string, any>;
   setObsGroupCounter?: any;
 }
--- a/src/registry/inbuilt-components/control-templates.ts
+++ b/src/registry/inbuilt-components/control-templates.ts
@@ -11,6 +11,12 @@ export const controlTemplates: Array<ControlTemplate> = [
     },
   },
   {
+    name: 'encounterProvider',
+    datasource: {
+      name: 'provider_datasource',
+    },
+  },
+  {
     name: 'problem',
     datasource: {
       name: 'problem_datasource',
--- a/src/registry/inbuilt-components/inbuiltControls.ts
+++ b/src/registry/inbuilt-components/inbuiltControls.ts
@@ -120,6 +120,6 @@ export const inbuiltControls: Array<RegistryItem<React.ComponentType<OHRIFormFie
   ...controlTemplates.map((template) => ({
     name: `${template.name}Control`,
     component: templateToComponentMap.find((component) => component.name === template.name).baseControlComponent,
-    type: template.name.toLowerCase(),
+    type: template.name,
   })),
 ];
--- a/src/registry/inbuilt-components/inbuiltDataSources.ts
+++ b/src/registry/inbuilt-components/inbuiltDataSources.ts
@@ -1,6 +1,7 @@
 import { DataSource } from '../../api/types';
 import { ConceptDataSource } from '../../datasources/concept-data-source';
 import { LocationDataSource } from '../../datasources/location-data-source';
+import { ProviderDataSource } from '../../datasources/provider-datasource';
 import { RegistryItem } from '../registry';
 
 /**
@@ -19,8 +20,12 @@ export const inbuiltDataSources: Array<RegistryItem<DataSource<any>>> = [
     name: 'problem_datasource',
     component: new ConceptDataSource(),
   },
+  {
+    name: 'provider_datasource',
+    component: new ProviderDataSource(),
+  },
 ];
 
 export const validateInbuiltDatasource = (name: string) => {
-  return inbuiltDataSources.some(datasource => datasource.name === name);
+  return inbuiltDataSources.some((datasource) => datasource.name === name);
 };
--- a/src/registry/inbuilt-components/inbuiltFieldSubmissionHandlers.ts
+++ b/src/registry/inbuilt-components/inbuiltFieldSubmissionHandlers.ts
@@ -1,6 +1,7 @@
 import { SubmissionHandler } from '../../api/types';
 import { ObsSubmissionHandler, EncounterLocationSubmissionHandler } from '../../submission-handlers/base-handlers';
 import { EncounterDatetimeHandler } from '../../submission-handlers/encounterDatetimeHandler';
+import { EncounterProviderHandler } from '../../submission-handlers/encounterProviderHandler';
 import { RegistryItem } from '../registry';
 
 /**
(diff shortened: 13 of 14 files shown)


src/components/section/helpers.ts
```diff
@@ -21,7 +21,9 @@ export function getFieldControlWithFallback(question: OHRIFormField) {
   if (question.type === 'encounterLocation') {
     question.questionOptions.rendering = 'encounter-location';
   }
-
+  if (question.type === 'encounterProvider') {
+    return getRegisteredControl('encounterProvider');
+  }
```
samuelmale, line 26: @kajambiya, there are two problematic issues with this:

1. We are gradually coupling the datatype property with the rendering property.
2. This function is performing more work than intended. Initially, it was meant to serve as a fallback to text rendering for fields lacking concepts. However, it now involves some form of mapping based on the datatype. Ideally, the `type` prop should configure the handler component, while the `rendering` should configure the actual control (the underlying component). Over time, this function may become bloated and challenging to maintain. Frankly, I don't see a good reason for this approach.

src/components/inputs/ui-select-extended/ui-select-extended.tsx
```diff
@@ -45,7 +45,10 @@ const UISelectExtended: React.FC<OHRIFormFieldProps> = ({ question, handler, onC
   }, [encounterContext.sessionMode, question.readonly, question.inlineRendering, layoutType, workspaceLayout]);
 
   useEffect(() => {
-    const datasourceName = question.questionOptions?.datasource?.name;
+    let datasourceName = question.questionOptions?.datasource?.name;
+    if (question.type === 'encounterProvider') {
+      datasourceName = question.type;
+    }
```
samuelmale, line 51: @kajambiya Any good reason for making this component aware of the `encounterProvider` type?

src/api/api.ts
```diff
@@ -71,6 +71,12 @@ export function getAllLocations(): Observable<{ uuid: string; display: string }[
   );
 }
 
+export function getEncounterProviders(): Observable<{ uuid: string; display: string }[]> {
+  return openmrsObservableFetch(`/ws/rest/v1/provider?v=custom:(uuid,display)`).pipe(
+    map(({ data }) => data['results']),
+  );
+}
```
samuelmale, line 78: Are you using this function?

samuelmale (commented):
Happy to see this taking shape, just a few early comments

src/api/api.ts
kajambiya: I had planned to use it my logic for getting the initial and previous value but that might change.

src/components/inputs/ui-select-extended/ui-select-extended.tsx
kajambiya: Same reason as highlighted above. Ideally if a control doesn't explicitly define a datasource, it will search for the datasource using the rendering (which will be the same name provided in the `controlTemplates`) but the AMPATH case makes it unique since they use type and not rendering to specify `encounterLocation`

src/components/section/helpers.ts
samuelmale: I understand the issue, @kajambiya. There are a couple of tweaks that could potentially solve it.

Let's lay all the pieces on the table: Our approach to supporting derived controls (components based on data sources plus the ui-select-extended component) relies on a combination of several elements:
 
-  **The ui-select-extended component (USE)**
 This is a generic component used to flesh out our derived control
- **A Datasource**
This is responsible for loading arbitrary data ie. coded answers from the backend 
- **A Submission handler**
This is responsible for form field value binding and a factory of composite field values

### How do we put these together?
At the heart of the form-engine is the registry, which manages all components of the form-engine. 
To get a derived control or rendering we [introduced control templates](https://github.com/openmrs/openmrs-form-engine-lib/blob/8c5cf5ba6df5180af0eb2603bd0ce3202ae2b8ec/src/registry/inbuilt-components/control-templates.ts#L3) that provide the default DS config, which are then used to [derive new controls](https://github.com/openmrs/openmrs-form-engine-lib/blob/8c5cf5ba6df5180af0eb2603bd0ce3202ae2b8ec/src/registry/inbuilt-components/inbuiltControls.ts#L120-L124). 

This approach isn't perfect but seems to work perfectly and keeps us operating within the premises of the Form Registry. 

### The simpler solution (Feels a bit hacky)
We don't have to worry about the rendering and submission handler because, at the end of the day, the correct components will be injected based on the config you shared above. However, we have to figure out how to inject the datasource along with the [default config](https://github.com/openmrs/openmrs-form-engine-lib/blob/8c5cf5ba6df5180af0eb2603bd0ce3202ae2b8ec/src/registry/inbuilt-components/control-templates.ts#L13-L25). Currently, if an instance of a USE component has no datasource config as part of the form field configuration, [we fallback to using the field rendering type](https://github.com/openmrs/openmrs-form-engine-lib/blob/8c5cf5ba6df5180af0eb2603bd0ce3202ae2b8ec/src/components/inputs/ui-select-extended/ui-select-extended.tsx#L49-L53) to identify an existing inbuilt DS template. The magic bullet is falling back to the `question.type` as opposed to the `question.questionOptions.rendering`. We just have to make sure that the control template name aligns with the value of the type prop. 
NB: This is a potential breaking change for rendering types such as "drug" and "problem" but easy to fix. 

### Second solution (I seem to prefer this one)
We can simply introduce a form schema adaptor or something along those lines. Its primary role could be to ensure that the schema is fine-tuned to what the renderer expects. Before rendering the form, we could perform some kind of schema validation and refinement.

Input:

```json
{
  "type": "encounterLocation",
  "label": "Facility name (site/satellite clinic required):",
  "id": "location",
  "required": "true",
  "questionOptions": {
    "rendering": "ui-select-extended"
  }
}
```

Output:

```json
{
  "type": "encounterLocation",
  "label": "Facility name (site/satellite clinic required):",
  "id": "location",
  "required": "true",
  "questionOptions": {
    "rendering": "encounter-location"
  }
}
```
Here we simply change the rendering to [encounter-location](https://github.com/openmrs/openmrs-form-engine-lib/blob/8c5cf5ba6df5180af0eb2603bd0ce3202ae2b8ec/src/registry/inbuilt-components/inbuiltControls.ts#L60-L64) (Note that this rendering is not based on the USE) we are currently doing something similar [here](https://github.com/openmrs/openmrs-form-engine-lib/blob/8c5cf5ba6df5180af0eb2603bd0ce3202ae2b8ec/src/components/section/helpers.ts#L22-L23). Thoughts? cc: @ibacher @denniskigen

kajambiya: Thanks @samuelmale 
I actually like both solutions. Let me wait to hear what other folks have to say

src/api/api.ts
kajambiya: Hi @samuelmale 
I've made some changes.
I haven't introduced a schema adaptor as you had earlier suggested. Realized I would have to first map through the entire schema(pages, sections, questions) to fine tune the schema and then I would have to map through the schema again to process sections. I therefore decided to add this transformation at the section level.
I've introduced the `transformSchema` function which checks every question in a section to see if it needs to be transformed. Let me know what you think.

denniskigen:
Very timely, thanks!

src/components/section/helpers.ts
```diff
@@ -44,3 +37,19 @@ export function hasMissingConcept(question: OHRIFormField) {
...
   );
 }
+
+/**
+ * Make schema transformations especially for originally AFE schemas to match the RFE schema
+ */
+export function transformSchema(question: OHRIFormField) {
+  switch (question.type) {
+    case 'encounterProvider':
+      question.questionOptions.rendering = 'encounterProvider';
```
eudson, line 47: very minor: @kajambiya apologies for my OCD but why do we have the rendering types following different cases? camelCase and kebab-case?

eudson: Also this function is not transforming the entire schema so I would suggest you rename it because it can be confusing and people can decide to add logic beyond just changing the field definitions

src/api/api.ts
eudson: @samuelmale thank  you so much for these fruitful comments which I agree. I think there is a lot of things that need improvement on this code base and perhaps we should add some or most of this type of comments (in case they were not resolved yet) to the concept note (links to the PRs) so that we can work on them when we in Kampala next month for the hackathon. 

@kajambiya I remember that in out tech session today we agreed that this `transformSchema` belongs to the entry-point of the engine where the initial transformations happen. Unless you can't do there because of the additional looping your are talking about then that is fine otherwise it will slowdown the engine.

src/api/types.ts
```diff
@@ -199,6 +199,7 @@ export type RenderType =
   | 'group'
   | 'content-switcher'
   | 'encounter-location'
+  | 'encounterProvider'
```
pirupius:
```suggestion
  | 'encounter-provider'
```

src/components/section/helpers.ts
```diff
@@ -42,3 +37,19 @@ export function hasMissingConcept(question: OHRIFormField) {
     question.type == 'obs' && !question.questionOptions.concept && question.questionOptions.rendering !== 'fixed-value'
   );
 }
+
+/**
+ * Make schema transformations especially for originally AFE schemas to match the RFE schema
+ */
+export function transformQuestion(question: OHRIFormField) {
```
pirupius, line 44: 🚀

src/components/section/ohri-form-section.component.tsx
```diff
@@ -111,7 +113,9 @@ const OHRIFormSection = ({ fields, onFieldChange }) => {
                         <PreviousValueReview
                           previousValue={previousFieldValue}
                           displayText={
-                            fieldDescriptor.questionOptions.rendering == 'checkbox'
+                            fieldDescriptor.questionOptions.rendering === 'date'
+                              ? formatDate(previousFieldValue)
+                              : fieldDescriptor.questionOptions.rendering == 'checkbox'
                               ? previousCheckboxDisplayValue
                               : previousFieldValue?.display
```
pirupius, line 120: This is getting overly nested and looks like something that might grow with more unique rendering types

src/datasources/provider-datasource.ts
```diff
@@ -0,0 +1,27 @@
+import { openmrsFetch } from '@openmrs/esm-framework';
+import { BaseOpenMRSDataSource } from './data-source';
+
+export class ProviderDataSource extends BaseOpenMRSDataSource {
+  constructor() {
+    super('/ws/rest/v1/provider?v=custom:(uuid,display)');
```
pirupius, line 6:
```suggestion
    super(`${restBaseUrl}/provider?v=custom:(uuid,display)`);
```

src/api/api.ts
samuelmale: >I remember that in out tech session today we agreed that this transformSchema belongs to the entry-point of the engine where the initial transformations happen.

I agree with you @eudson and the best place I can think of is [here](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/hooks/useFormJson.tsx#L59)

src/datasources/provider-datasource.ts
kajambiya: The file has only 1 single usage of the url

pirupius: It's not about the usages. It's a pattern that has happened in all other monorepos to use `restBaseUrl` for the first potion of the rest api url.

kajambiya: Done

src/api/api.ts
kajambiya: Done

src/components/section/helpers.ts
```diff
@@ -42,3 +38,17 @@ export function hasMissingConcept(question: OHRIFormField) {
...
+
+export const formatPreviousValueDisplayText = (question: OHRIFormField, previousFieldValue: any) => {
+  if (question.questionOptions.rendering === 'date') {
+    return formatDate(previousFieldValue);
+  }
+  if (question.questionOptions.rendering === 'checkbox') {
+    return Array.isArray(previousFieldValue) ? previousValueDisplayForCheckbox(previousFieldValue) : null;
+  }
+  return previousFieldValue?.display;
+};
```
samuelmale, line 54:
```suggestion
export const formatFieldValueDisplay = (question: OHRIFormField, value: any) => {
  switch (question.questionOptions.rendering) {
    case 'date':
      return formatDate(value);
    case 'checkbox':
      return Array.isArray(value) ? value.map((entry) => entry['display']).join(', ') : value;
    default:
      return value?.display ?? value;
  }
};
```

src/datasources/provider-datasource.ts
```diff
@@ -0,0 +1,27 @@
...
+  fetchData(searchTerm: string, config?: Record<string, any>, uuid?: string): Promise<any[]> {
+    let apiUrl = this.url;
+    const urlParts = apiUrl.split('?');
+    if (config?.tag) {
+      apiUrl = `${urlParts[0]}?tag=${config.tag}&${urlParts[1]}`;
+    }
+    //overwrite url if there's a uuid value, meaning we are in edit mode
+    if (uuid) {
+      apiUrl = `${urlParts[0]}/${uuid}?${urlParts[1]}`;
+    }
```
samuelmale, line 18: What are you trying to achieve here?

src/hooks/useFormJson.tsx
```diff
@@ -181,6 +184,38 @@ function updateFormJsonWithComponents(formJson: OHRIFormSchema, formComponents:
...
+        if (section.questions) {
+          let questions = section.questions;
+          questions.map((question) => {
+            transformQuestion(question);
+          });
+        }
+      });
+    }
+  });
+};
```
samuelmale, line 201: This should be move to its own file and should be configurable. Transformers could implement some special interface:

```ts

interface FormSchemaTransformer {
   transform: (form: OHRIFormSchema) => OHRIFormSchema;
}

// AFE transformer

const AngularFormTransformer: FormSchemaTransformer = {
  transform: (form) => {
    // do transformations
    return form; 
  }
};

// transformer-form-schema.ts

function transformSchema(form: OHRIFormSchema) {
  const transformers = getRegisteredTransformers();
  transformers.forEach(transformer => {
    form = transformer.transform(form);
  });
  return form;
}
```

src/datasources/provider-datasource.ts
kajambiya: This was a technical debt on the `ui-select-extended`
Remember we had the default one and the `searchable`.
There was an issue with the `searchable` one in edit mode - The saved value would't be picked because at the moment items would be empty since there's no attempt to search yet. However, the selected item was being picked as below:
`selectedItem={items.find((item) => item.uuid == field.value)}`
The solution was to add an optional parameter(uuid) to the `fetchData` function so that if the `ui-select-extended` is searchable, in edit mode, we pass the saved value(uuid) to the `fetchData` function so that we can pick the corresponding display value.

Not sure whether my explanation is not confusing

samuelmale: Yes it makes sense to me, not a perfect way of solving this problem but it works.

src/components/section/helpers.ts
kajambiya: Done

src/hooks/useFormJson.tsx
kajambiya: Done

src/datasources/provider-datasource.ts
kajambiya: Agree.
I think we can brainstorm on the most perfect solution during the hackathon

src/registry/registry.ts
```diff
@@ -123,6 +142,48 @@ export async function getRegisteredFieldSubmissionHandler(type: string): Promise
...
+      transformers.push(inbuiltTransformer.component);
+    }
+  }
+
+  // Cache all fetched transformers
+  for (const transformer of transformers) {
+    registryCache.formTransformers[
+      inbuiltFormTransformers.find((inbuiltTransformer) => inbuiltTransformer.component === transformer).name
+    ] = transformer;
+  }
```
samuelmale, line 182: I don't understand why this logic is this complicated. What makes transformers special from other registry components? As an example here is a similar implementation for field handlers:

```ts
export async function getRegisteredFieldSubmissionHandler(type: string): Promise<SubmissionHandler> {
  if (registryCache.fieldSubmissionHandlers[type]) {
    return registryCache.fieldSubmissionHandlers[type];
  }
  let handler = inbuiltFieldSubmissionHandlers.find((handler) => handler.type === type)?.component;
  // if undefined, try serching through the registered custom handlers
  if (!handler) {
    const handlerImport = await getFormsStore()
      .fieldSubmissionHandlers.find((handler) => handler.type === type)
      ?.load?.();
    handler = handlerImport?.default;
  }
  registryCache.fieldSubmissionHandlers[type] = handler;
  return handler;
}
```

src/api/types.ts
```diff
@@ -332,3 +333,10 @@ export interface ProgramEnrollmentPayload {
   dateCompleted?: string;
   location: string;
 }
+
+/**
+ * Defines logic that transforms the form schema
+ */
+export interface FormSchemaTransformer {
+  transform: (form: OHRIFormSchema) => void;
+}
```
samuelmale, line 342:
```suggestion

/**
 * A form schema transformer is used to bridge the gap caused by different variations of form schemas
 * in the OpenMRS JSON schema-based form-entry world. It fine-tunes custom schemas to be compliant
 * with the React Form Engine.
 */
export interface FormSchemaTransformer {
  /**
   * Transforms the raw schema to be compatible with the React Form Engine.
   */
  transform: (rawSchema: OHRIFormSchema) => OHRIFormSchema;
}
```

src/datasources/provider-datasource.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    return openmrsFetch(searchTerm ? `${apiUrl}&q=${searchTerm}` : apiUrl).then(({ data }) => {
+      if (data.results) {
+        return data.results;
+      }
+      return data;
+    });
+  }
```
samuelmale, line 26: This implementation is unnecessarily complicated likely because of a poor abstraction in the `BaseOpenMRSDataSource`. I also see copy and paste residue eg. the part where you're applying a "tag"; providers don't support Tags. I would refactor to simply:

```ts
export class ProviderDataSource extends BaseOpenMRSDataSource {
  constructor() {
    super(null);
  }

  async fetchData(searchTerm: string, config?: Record<string, any>, uuid?: string): Promise<any[]> {
    const rep = 'v=custom:(uuid,display)';
    const url = uuid ? `${restBaseUrl}/provider/${uuid}?${rep}` : `${restBaseUrl}/provider?${rep}`;
    const { data } = await openmrsFetch(searchTerm ? `${url}&q=${searchTerm}` : url);
    if (data.results) {
      return data.results;
    }
    return Array.isArray(data) ? data : [data];
  }
}
```

src/registry/inbuilt-components/inbuiltTransformers.ts
```diff
@@ -0,0 +1,10 @@
+import { FormSchemaTransformer } from '../../api/types';
+import { AngularFormTransformer } from '../../transformers/AFETransformer';
+import { RegistryItem } from '../registry';
+
+export const inbuiltFormTransformers: Array<RegistryItem<FormSchemaTransformer>> = [
+  {
+    name: 'AFESChemaTransformer',
+    component: AngularFormTransformer,
+  },
```
samuelmale, line 9: Nit:

```ts
 {
    name: 'AngularFormEngineSchemaTransformer',
    component: AngularFormEngineSchemaTransformer,
 },
```

src/transformers/AFETransformer.ts
```diff
@@ -0,0 +1,35 @@
+import { FormSchemaTransformer, OHRIFormField, OHRIFormSchema } from '../api/types';
+
+export const AngularFormTransformer: FormSchemaTransformer = {
```
samuelmale, line 3: I would rename to something like: `AngularFormEngineSchemaTransformer` and also rename the file to: `angular-fe-schema-transformer.ts`

src/transformers/form-schema-transformer.ts
```diff
@@ -0,0 +1,9 @@
+import { OHRIFormSchema } from '../api/types';
+import { getRegisteredformSchemaTransformers } from '../registry/registry';
+
+export async function transformSchema(form: OHRIFormSchema) {
+  const transformers = await getRegisteredformSchemaTransformers();
+  transformers.forEach((transformer) => {
+    transformer.transform(form);
+  });
```
samuelmale, line 8: I would use a reducer:

```ts
const refinedSchema = transformers.reduce((schema, transformer) => {
    return transformer.transform(schema);
  }, form);
  
return refinedSchema;
```

samuelmale (changes requested):
I'm starting to think we need to rethink the forms registry implementation. We should consider powering the registry by an IoC container like [inversify](https://github.com/inversify/InversifyJS). Otherwise thanks for the great work @kajambiya, just a few final comments.

src/registry/registry.ts
kajambiya: From your guidance earlier, we had to get all the transformers at once.
The difference with  `getRegisteredFieldSubmissionHandler` is that it only returns one handler at a time

samuelmale: I could be wrong here but getting all transformers in one shot can be like "Hey registry store, give me all you got". Something like:

```ts

const cache = [];

async function getRegisteredTransformers() {
  if (cache.length) {
    return cache;
  }

  const customTransformers = await Promise.all(
    formTransformersFromStore.map(async (registration) => {
      try {
        const transformerImport = await registration.load?.();
        return transformerImport?.default;
      } catch (error) {
        console.error("Error loading custom transformer:", error);
        return null; // Handle loading error gracefully
      }
    })
  );
  const allTransformers = [...inbuiltFormTransformers.map((registration) => registration.component), ...customTransformers.filter(Boolean)]; // Filter out null values

  // Cache the transformers for future use
  cache.push(...allTransformers);

  return allTransformers;
}
```

kajambiya:
> I'm starting to believe that we need to rethink the forms registry implementation. We could consider powering the registry with an IoC container like [inversify](https://github.com/inversify/InversifyJS). Otherwise thanks for the great work @kajambiya, below are my final comments.

Hi @samuelmale 
I've addressed the issues save for the change requested on the `getRegisteredFieldSubmissionHandler` function for which I need more clarity. I've also included the logic for loading default values for `encounterProvider, encounterLocation & encounterDatetime.` Please review and revert.

src/registry/registry.ts
samuelmale: @kajambiya did you take action on the above?

src/submission-handlers/encounterProviderHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+  },
+  getInitialValue: (
+    encounter: OpenmrsEncounter,
+    field: OHRIFormField,
+    allFormFields: Array<OHRIFormField>,
+    context: EncounterContext,
+  ) => {
+    if (encounter) {
+      const encounterProvider = encounter.encounterProviders[0]?.provider;
+      return encounterProvider ? encounterProvider.uuid : null;
```
samuelmale, line 18:
```suggestion
      return encounter.encounterProviders[0]?.provider?.uuid;
      
```

src/submission-handlers/encounterProviderHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+    } else {
+      return context.encounterProvider;
+    }
+  },
+
+  getDisplayValue: (field: OHRIFormField, value: any) => {
+    if (!field.value) {
+      return null;
+    }
+    return value;
```
samuelmale, line 28:
```suggestion 
    return field.value;
```

src/hooks/useInitialValues.ts
```diff
@@ -18,6 +18,7 @@ export function useInitialValues(
   const [isEncounterBindingComplete, setIsEncounterBindingComplete] = useState(
     encounterContext.sessionMode === 'enter',
   );
+  const sessionInferredValueTypes = ['encounterProvider', 'encounterDatetime', 'encounterLocation'];
```
samuelmale, line 21:
```suggestion
  const encounterContextInitialisableTypes = ['encounterProvider', 'encounterDatetime', 'encounterLocation'];
```

src/registry/registry.ts
```diff
@@ -83,6 +98,18 @@ export function registerExpressionHelper(name: string, fn: Function) {
   getFormsStore().expressionHelpers[name] = fn;
 }
 
+export function registereformSchemaTransformers(registration: FormScemaTransformerRegistration) {
```
samuelmale, line 101:
```suggestion
export function registerformSchemaTransformer(registration: ComponentRegistration<FormSchemaTransformer>) {
```

src/registry/registry.ts
```diff
@@ -83,6 +98,18 @@ export function registerExpressionHelper(name: string, fn: Function) {
...
+  const store = getFormsStore();
+  const existingIndex = store.formTransformers.findIndex((reg) => reg.name === registration.name);
+
+  if (existingIndex !== -1) {
+    // If registration with the same name exists, override it
+    store.formTransformers[existingIndex] = registration;
+  } else {
+    store.formTransformers.push(registration);
+  }
+}
```
samuelmale, line 111: This seems to imply that the transformers are stored in array yet they seem to be stored in a map. I would simply:

```ts
getFormsStore().formTransformers[registration.name] = registration;
```

src/registry/registry.ts
```diff
@@ -32,13 +40,18 @@ export interface FieldSubmissionHandlerRegistration extends ComponentRegistratio
   type: string;
 }
 
+export interface FormScemaTransformerRegistration extends ComponentRegistration<FormSchemaTransformer> {
+  type: string;
+}
```
samuelmale, line 45: There are two issues:

1. There is a typo in the interface name
2. Do we need the `type` property?

src/registry/registry.ts
kajambiya: Hi @samuelmale 
If I got you right, it's actually an array.
This is how the FormsRegistryStoreState looks like:
```
  export interface FormsRegistryStoreState {
  controls: CustomControlRegistration[];
  fieldValidators: ComponentRegistration<FieldValidator>[];
  fieldSubmissionHandlers: FieldSubmissionHandlerRegistration[];
  postSubmissionActions: ComponentRegistration<PostSubmissionAction>[];
  dataSources: ComponentRegistration<DataSource<any>>[];
  expressionHelpers: Record<string, Function>;
  formSchemaTransformers: FormSchemaTransformerRegistration[];
}
```

I followed the same logic for submission handlers, the only difference being that for transformers we override
```
export function registerFieldSubmissionHandler(registration: FieldSubmissionHandlerRegistration) {
  getFormsStore().fieldSubmissionHandlers.push(registration);
}
```

samuelmale (approved):

eudson:
Thank you @samuelmale

PR #177, opened 2024-04-05
fix(O3-2167): Add meta property to OHRIFormField interface

senthil-k8s:
## Requirements

## Summary
Storing the Concept object in OHRIFormfield, in order to reduce the db calls to fetch concept information whenever needed.

**Ways for Storing the Concept Object:**

**Option 1:** Modify "concept" field: Change the type of the "concept" field from a string to an object (or allow it to hold both strings and objects). This approach directly replaces the existing concept storage mechanism.

**Option 2:** Add new "meta" property: Introduce a new property named "meta" (type: Record<string, any>) within the data structure. This "meta" property would act as a dictionary to store additional information, including the full concept object under a specific key.

**Preferred:** Adding a new "meta" property seems like a more suitable approach for the following reasons:
**Backward Compatibility:** It maintains compatibility with existing code that uses the "concept" field for concept UUIDs or mappings.
**Flexibility:** The "meta" property can be used to store other relevant information about the question beyond just the concept object.
**Clarity:** It keeps the core data structure cleaner by separating the concept object from core question properties like label and type.

## Screenshots
[Screen-Recorder](https://www.loom.com/share/b1a0b058026447e193165ab7be4acddd?sid=4fa6b59c-0f89-441e-8ac2-44dc9214b5af)

## Related Issue
[O3-2167](https://openmrs.atlassian.net/browse/O3-2167)


samuelmale:
Hi @senthil-k8s I seem to like the `meta` prop. Are you planning on actually using this property to address the issue here: https://openmrs.atlassian.net/browse/O3-2167?

samuelmale:
Yes that's clear enough and it makes sense @senthil-k8s. So are you planning on consuming the meta prop?

senthil-k8s:
Yes, @samuelmale , let's say in the future if we need concept information (it could be datatypes or classes), we no longer need to fetch the data every time; instead, we can get it from the meta prop. your suggestions are also welcome @samuelmale

samuelmale:
@senthil-k8s this PR just introduces a meta prop to the `OHRIFormField` interface and linting. In as much as this maybe necessary, it doesn't address the issue defined in the ticket. If the ticket isn't clear enough to you, feel free to claim a different task

PR #215, opened 2024-04-24
(feat) Setup i18n for form engine lib

vasharma05:
## Requirements

## Summary
This PR implements internationalization in the Form Engine Library. Whenever the form engine lib is loaded, the lib will look for the active locale via the `window.i18next.language` and then load the particular file from the translation folder into the i18next instance. The same instance is then shared with the OHRIForm using the `I18NextProvider` via a HOC `withTranslation`.

Please note that in case the language is loaded, the corresponding translation resource will be loaded into the i18next in the runtime.

## Screenshots
image

image

## Related Issue
TBD


src/i18n.ts
```diff
@@ -0,0 +1,23 @@
+function loadResourcesFromFile() {
+  const lang = window.i18next.language;
```
ibacher, line 2: Rather than directly referencing `window.i18next`, I'd mark `i18next` as a peer dependency and import it in this file.

src/i18n.ts
```diff
@@ -0,0 +1,23 @@
...
+  import(`../translations/${lang}.json`)
```
ibacher, line 3:
```suggestion
  import(/* webpackMode: "lazy" */ `../translations/${lang}.json`)
```

src/i18n.ts
```diff
@@ -0,0 +1,23 @@
...
+      window.i18next.addResourceBundle(lang, '@openmrs/openmrs-form-engine-lib', data);
+    })
+    .catch((err) => console.error(err));
+}
+
+function setupFormEngineLibI18n() {
+  loadResourcesFromFile();
+  window.i18next.on('languageChanged', loadResourcesFromFile);
+
+  const languageChangeObserver = new MutationObserver(loadResourcesFromFile);
```
ibacher, line 15: I'd stick to just binding to the `i18next` `onLanguageChanged` event.

src/i18n.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  languageChangeObserver.observe(document.documentElement, {
+    attributeFilter: ['lang'],
+    attributes: true,
+  });
+}
+
+setupFormEngineLibI18n();
```
ibacher, line 23: Instead of having this implicitly loaded, this should be explicitly called as part of the form engine initialization.

translations/fr.json
```diff
@@ -0,0 +1,39 @@
+{
```
ibacher, line 1: These translations should be added via Transifex.

src/withTranslation.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { I18nextProvider } from 'react-i18next';
+
+export default function withTranslation(WrappedComponent: React.ComponentType<any>) {
+  return function WithTranslation(props: any) {
+    return (
+      <I18nextProvider i18n={window.i18next} defaultNS="@openmrs/openmrs-form-engine-lib">
```
ibacher, line 7: We need to be careful here: it is appropriate for `@openmrs/openmrs-form-engine-lib` to be used as the namespace for strings defined in the _form engine library_, but _form specific translations_ (like concepts and labels) should be confined to a form-specific namespace.

tsconfig.json
```diff
@@ -6,19 +6,10 @@
...
-      "es2018",
-      "es2020"
-    ],
+    "lib": ["dom", "es5", "scripthost", "es2015", "es2015.promise", "es2016.array.include", "es2018", "es2020"],
     "resolveJsonModule": true,
     "noEmit": true,
     "target": "esnext"
   },
-  "include": ["src/**/*", "./setupTests.ts"]
+  "include": ["src/**/*", "./setupTests.ts", "tools/i18next-parser.config.js", "prettier.config.js"]
```
ibacher, line 14: I don't think these should be mentioned in `tsconfig` at all? They aren't even Typescript...

translations/fr.json
denniskigen: For non-en locales.

src/i18n.ts
vasharma05: Getting undefined with `i18next` reference. Do you know what might be the reason?

denniskigen:
window.18next.on is breaking the tests. Do we have a mock for it somewhere you're aware of, @ibacher?

src/withTranslation.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { I18nextProvider } from 'react-i18next';
+
+export default function withTranslation(WrappedComponent: React.ComponentType<any>) {
```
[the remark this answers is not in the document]
vasharma05: Hi @ibacher !
Actually the `withTranslation` from `i18next` injects the `t` and `i18next` in the Component's props, but it doesn't set the context of the `i18next` instance and the provided namespace, hence when using `useTranslation`, it takes the defaultNamespace as `translation` and new `i18n` instance.

I tested it out `withTranslation` from `i18next` and it didn't work. We need to share the namespace via the Provider for it to be applicable whole form wide.

src/withTranslation.tsx
vasharma05: Hi @ibacher!
I understand that this is to be done not to conflict the common translation keys between the forms, but everytime we load a form, if we override the form's translations in the previously presented translations, I believe we can skip having and handling namespaces based on forms. Hence saving the effort of handling translations for 2 kinds: 1. Static text translations and 2. Form translations.

What do you think?

ibacher:
@denniskigen No, but I _think_ my suggestion of using `i18next` as a shared library will render it unnecessary.

src/withTranslation.tsx
ibacher: @vasharma05 I'm ok with not having a namespace-per-individual-form, but I do think we need to distinguish between the namespace for translating things inside the form-engine-lib and the namespace for translating form-specific stuff, if only because the former only needs to be loaded once, no?

src/withTranslation.tsx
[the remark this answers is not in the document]
ibacher: I yes, that's a good point. The context isn't really set for things further down the tree.

translations/fr.json
vasharma05: Was added for testing purposes, will revert it.
Thanks!

src/withTranslation.tsx
vasharma05: @ibacher , I will not be adding the static translations again, I will just be adding the form translations in the existing and loaded resource. I referred this from Ngx Form entry.

src/i18n.ts
ibacher: When in the app is that happening? It's vaguely possible that `language` is `undefined`, but `window.i18next` should be set [here](https://github.com/openmrs/openmrs-esm-core/blob/eeb48a4c573111d7323956164d75e7f64723543b/packages/shell/esm-app-shell/src/locale.ts#L14). That's called _after_ apps are registered, but it should be before anything gets mounted (since single-spa isn't called until `setupI18n()` succeeds).

vasharma05:
Hi @ibacher!
I have added another set of changes, please review.

image

src/i18n.ts
vasharma05: image
image

src/withFormEngineTranslations.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export default function withFormEngineTranslations(WrappedComponent: React.ComponentType<any>) {
+  return function WithTranslation(props: any) {
+    return (
+      <I18nextProvider i18n={window.i18next} defaultNS="@openmrs/openmrs-form-engine-lib">
+        <WrappedComponent {...props} />
+      </I18nextProvider>
+    );
+  };
+}
```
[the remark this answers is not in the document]
vasharma05: Should be resolved now, add the mock for I18nextProvider

denniskigen (approved):
Really excellent work, @vasharma05!

denniskigen (approved):
Really excellent work, @vasharma05!

denniskigen (approved):
Really excellent work, @vasharma05!

ibacher (approved):
I wonder if some of the issue is that we're using i18next 21.x rather than 23.x? Anyways, very nice work!

PR #250, opened 2024-04-30
(fix) Fix layout shifts caused by loaders

denniskigen:
## Requirements

## Summary

Several loaders in the library currently cause layout shifts. This PR targets two of them:

- The linear loader that gets shown when loading the form dependencies (specifically the related encounter). More important than showing a loading state is rendering an error message if the encounter fails to load. I've elected to remove the loader.
- The spinner shown when loading data sources for the `UiSelectExtended` component. I've tweaked the styling on its parent container so that the spinner is placed to the right of the dropdown in a flex container and centre aligned so it matches up to the dropdown's height. 

### Before

CleanShot 2024-04-30 at 7  37 48@2x

### After

CleanShot 2024-05-02 at 11  32 02@2x


src/components/inputs/content-switcher/content-switcher.test.tsx
```diff
@@ -51,9 +51,9 @@ const encounterContext: EncounterContext = {
   setEncounterLocation: jest.fn,
 };
 
-const renderForm = (intialValues) => {
+const renderForm = (initialValues) => {
```
denniskigen, line 54: Just renames the variable.

src/form-engine.component.tsx
```diff
@@ -255,11 +254,6 @@ const FormEngine: React.FC<FormProps> = ({
               <LoadingIcon />
             ) : (
               <div className={styles.formEngineContainer}>
-                {isLoadingFormDependencies && (
-                  <div className={styles.loader}>
-                    <LinearLoader />
-                  </div>
-                )}
```
samuelmale, line 262: @denniskigen this renders the `isLoadingFormDependencies` state redundant. We either remove or use it. I think we still need some kind of loader for form dependencies. I've no objection with moving this to the standard loader.

denniskigen: Sure. So I've incorporated `LinearLoader` into the `FormEngine` component in my latest commits. I've also replaced the `UiSelectExtended` component's loader with a [DropdownSkeleton](https://react.carbondesignsystem.com/?path=/story/components-dropdown--skeleton).

src/components/loaders/loader.component.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { InlineLoading } from '@carbon/react';
+import styles from './loader.scss';
+
+const isLoadingPreviousEncounter: React.FC = () => {
```
samuelmale, line 6: This name is kind of confusing, someone may think this is some React state that flags whether we are loading the `previousEncounter`; yet it's the loader itself. I would rename it to something like `Loader`.

samuelmale (approved):
LGTM, thanks @denniskigen 🍻

src/components/loaders/loader.component.tsx
denniskigen: Nice catch 😅

PR #256, opened 2024-05-06
field-validation:Enabling default field validation for multi-checkbox field values 

gitcliff:
## Requirements

## Summary
The [default field validation](https://github.com/openmrs/openmrs-form-engine-lib/blob/40ec1b3fcd3132735fe8d8840aed851d3d1f740b/src/validators/default-value-validator.ts#L7-L14) is only looking at the scenario when the value is an array of one element , however for the case of the checkbox and multi-checbox the value will have more than one element ie  ['30e2da8f-34ca-4c93-94c8-d429f22d381c', '1e2918fb-4e15-11e4-8a57-0800271c1b75'] which causes Value not found in coded answers list error to be displayed as seen in the screenshot below yet both coded values are present in the answers coded list. We need to cater for multi-select fields to
## Screenshots
Screenshot 2024-05-06 at 11 27 21 PM

### DIFF
--- a/src/validators/default-value-validator.ts
+++ b/src/validators/default-value-validator.ts
@@ -5,7 +5,11 @@ import { codedTypes } from '../constants';
 export const DefaultFieldValueValidator: FormFieldValidator = {
   validate: (field: FormField, value: any) => {
     if (codedTypes.includes(field.questionOptions.rendering)) {
-      if (!value.some((val) => field.questionOptions.answers?.find((answer) => answer.concept === val))) {
+      const valuesToCheck = Array.isArray(value) ? value : [value];
+      // check whether value exists in answers
+      if (
+        !valuesToCheck.some((val: string) => field.questionOptions.answers?.find((answer) => answer.concept === val))
+      ) {
         return [
           { resultType: 'error', errCode: 'invalid.defaultValue', message: 'Value not found in coded answers list' },
         ];
--- a/src/validators/default-value-validator.ts
+++ b/src/validators/default-value-validator.ts
@@ -8,7 +8,7 @@ export const DefaultFieldValueValidator: FormFieldValidator = {
       const valuesToCheck = Array.isArray(value) ? value : [value];
       // check whether value exists in answers
       if (
-        !valuesToCheck.some((val: string) => field.questionOptions.answers?.find((answer) => answer.concept === val))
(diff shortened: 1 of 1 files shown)


src/validators/default-value-validator.ts
```diff
@@ -5,8 +5,11 @@ import { codedTypes } from '../constants';
 export const DefaultFieldValueValidator: FormFieldValidator = {
   validate: (field: FormField, value: any) => {
     if (codedTypes.includes(field.questionOptions.rendering)) {
+      const valuesToCheck = Array.isArray(value) ? value : [value];
```
samuelmale, line 8: I would name this to something like: `valuesArray`

src/validators/default-value-validator.ts
```diff
@@ -5,8 +5,11 @@ import { codedTypes } from '../constants';
...
       // check whether value exists in answers
-      if (!field.questionOptions.answers?.find((answer) => answer.concept == value)) {
+      if (
+        !valuesToCheck.some((val: string) => field.questionOptions.answers?.find((answer) => answer.concept === val))
```
samuelmale, line 11: `some(...)` tests whether at least one element passes the condition specified in the callback, you want to use `every(...)`

```ts
!valuesArray.every((value: string) => field.questionOptions.answers.includes(value));
```

samuelmale (changes requested):

src/validators/default-value-validator.ts
gitcliff: @samuelmale switched to` every(...)` and maintained the callback condition to check if for every val in `valuesToCheck`, there exists at least one answer in `field.questionOptions.answers` where `answer.concept` is equal to `val`

samuelmale: @gitcliff any reason as to why you're using `find` as opposed to using `includes`?

```ts
!valuesArray.every((value: string) => field.questionOptions.answers.includes(value));
```

gitcliff: @samuelmale when i use includes the validation fails , secondly the elements in the `valuesArray` are concepts UUIDs. 
field.questionOptions.answers array holds ie```
 {
                        "concept": "30e2da8f-34ca-4c93-94c8-d429f22d381c",
                        "label": "Congestion"
                      }
```

samuelmale (approved):

PR #258, opened 2024-05-07
(fix) Multi-select component to initialise and display consistently in edit mode

samuelmale:
## Requirements

## Summary

This aims to fix an issue with the multi-select component where it doesn't preselect previously selected values while in edit mode. 

## Screenshots

https://www.loom.com/share/6a7b2ac48ff7430ba25e6a75da42c1d3

## Related Issue

https://openmrs.atlassian.net/browse/O3-3135

### DIFF
--- a/src/components/inputs/multi-select/multi-select.component.tsx
+++ b/src/components/inputs/multi-select/multi-select.component.tsx
@@ -16,24 +16,12 @@ import styles from './multi-select.scss';
 const MultiSelect: React.FC<FormFieldProps> = ({ question, onChange, handler, previousValue }) => {
   const { t } = useTranslation();
   const [field, meta] = useField(question.id);
-  const { setFieldValue, encounterContext, layoutType, workspaceLayout } = React.useContext(FormContext);
+  const { setFieldValue, encounterContext, layoutType, workspaceLayout, isFieldInitializationComplete } =
+    React.useContext(FormContext);
   const [errors, setErrors] = useState([]);
   const [warnings, setWarnings] = useState([]);
   const [counter, setCounter] = useState(0);
-  const [touched, setTouched] = useState(false);
   const isFieldRequiredError = useMemo(() => errors[0]?.errCode == fieldRequiredErrCode, [errors]);
-
-  useEffect(() => {
-    // Carbon's MultiSelect has issues related to not updating the component based on the `initialSelectedItems` prop
-    // this is an intermittent issue. As a temporary solution, were are forcing the component to re-render
-    if (field.value && field.value.length == 0) {
-      // chances are high the value was cleared
-      // force the Multiselect component to be re-mounted
-      setCounter(counter + 1);
-    } else if (!touched && question.value) {
-      setCounter(counter + 1);
-    }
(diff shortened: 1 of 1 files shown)


gitcliff:
thanks @samuelmale LGTM

src/components/inputs/multi-select/multi-select.component.tsx
```diff
@@ -16,44 +16,39 @@ import styles from './multi-select.scss';
...
 
+  const selectOptions = question.questionOptions.answers
+    .filter((answer) => !answer.isHidden)
+    .map((answer, index) => ({
+      id: `${question.id}-${answer.concept}`,
+      concept: answer.concept,
+      label: answer.label,
+      key: index,
+      disabled: answer.disable?.isDisabled,
+    }));
```
vasharma05, line 41: Can we memoise this?

src/components/inputs/multi-select/multi-select.component.tsx
```diff
@@ -16,44 +16,39 @@ import styles from './multi-select.scss';
...
-      }
-    });
-    return selectedItems;
-  }, [question, field.value]);
+    if (isFieldInitializationComplete && field.value?.length && counter < 1) {
+      setCounter(counter + 1);
+      return selectOptions.filter((item) => field.value?.includes(item.concept));
+    }
+    return [];
+  }, [isFieldInitializationComplete, field.value]);
```
vasharma05, line 49: `counter`, `setCounter` and `selectOptions` will also be a dependency.

vasharma05 (approved):
A few changes, else looks good to me.
Thanks!

src/components/inputs/multi-select/multi-select.component.tsx
samuelmale: In my first approach memoised it but for some reason (will be addressed during the formik migration), some state events weren't properly being propagated. An example is the `answer.disable. isDisabled`. This property is computable based on skip logic, when I have it like:

```ts
const selectOptions = useMemo(
    () =>
      question.questionOptions.answers
        .filter((answer) => !answer.isHidden)
        .map((answer, index) => ({
          id: `${question.id}-${answer.concept}`,
          concept: answer.concept,
          label: answer.label,
          key: index,
          disabled: answer.disable?.isDisabled,
        })),
    [question],
  );

```

The new state computation of `answer.disable?.isDisabled` doesn't get propagated after skip logic runs. I guess it's the same reason why it wasn't [memoised originally](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/components/inputs/multi-select/multi-select.component.tsx#L99-L107).

src/components/inputs/multi-select/multi-select.component.tsx
samuelmale: Well yes but we don't need to subscribe to them to evaluate the `initiallySelectedQuestionItems`.
1. Subscribing to `selectOptions` introduces a weird bug (I had it as a dep in my initial implementation). 
2. We evaluate the state only if the `counter` is 0 so we don't have to subscribe to it.

PR #241, opened 2024-04-26
(feat) Fetch all concept references of the the form with SWR Infinite

vasharma05:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary
Currently, out of all concept references, only the top 50 references are fetched. With `useSWRInfinite` we can fetch all the concept references for the form in batches.

## Related Issue
None

PR #259, opened 2024-05-07
(fix) O3-2168: Concept reference chunk size depends on REST `webservices.rest.maxResultsCount` property

vasharma05:
## Requirements

## Summary

The count of concepts in a form can exceed the maximum results returned by the REST module. Hence, we use the global property `webservices.rest.maxResultsCount` system setting, defining the maximum results count the REST module can deliver in a single result.

Hence we create the chunks of UUIDs in the size of the value returned by the above mentioned system settings and fetch the results accordingly.

### Test fix
This PR reverts the change made in **form-helper** in #241. 

For the URLs where the response is not mocked, the `openmrsFetch` returns `undefined` in the tests, which when read by `useSWRInfinite` maps the response to `[ undefined ]`, which then breaks the function written in the form-helper (link mentioned above).

```
Error: Cannot read `uuid` of undefined
```

## Screenshots
None

## Related Issue
None

### DIFF
--- a/src/hooks/useConcepts.tsx
+++ b/src/hooks/useConcepts.tsx
@@ -1,22 +1,31 @@
 import { useMemo } from 'react';
 import { type FetchResponse, type OpenmrsResource, openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
 import useSWRInfinite from 'swr/infinite';
+import { useSystemSetting } from './useSystemSetting';
+
+type ConceptFetchResponse = FetchResponse<{ results: Array<OpenmrsResource> }>;
 
 const conceptRepresentation =
   'custom:(uuid,display,conceptClass:(uuid,display),answers:(uuid,display),conceptMappings:(conceptReferenceTerm:(conceptSource:(name),code)))';
-
-const chunkSize = 100;
 
 export function useConcepts(references: Set<string>): {
   concepts: Array<OpenmrsResource> | undefined;
   isLoading: boolean;
   error: Error | undefined;
 } {
+  const { isLoading: isLoadingMaxResultsDefault, systemSetting } = useSystemSetting(
+    'webservices.rest.maxResultsDefault',
+  );
+  const chunkSize = systemSetting?.value ? parseInt(systemSetting.value) : null;
   const totalCount = references.size;
   const totalPages = Math.ceil(totalCount / chunkSize);
 
-  const getUrl = (index) => {
+  const getUrl = (index, prevPageData: ConceptFetchResponse) => {
     if (index >= totalPages) {
+      return null;
+    }
+
+    if (!chunkSize) {
       return null;
     }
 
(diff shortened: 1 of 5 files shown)


src/hooks/useConcepts.tsx
```diff
@@ -39,7 +39,10 @@ export function useConcepts(references: Set<string>): {
 
   const results = useMemo(
     () => ({
-      concepts: data ? [].concat(data?.map((res) => res?.data?.results).flat()) : undefined,
+      // data?.[0] check is added for tests, as response is undefined in tests
+      // hence the returned concepts are [undefined], which breaks the form-helper.ts
+      // As it cannot read `uuid` of `undefined`
+      concepts: data && data?.[0] ? [].concat(...data.map((res) => res?.data?.results)) : undefined,
```
samuelmale, line 45:
```suggestion
      concepts: data?.[0] ? [].concat(data.map((res) => res?.data?.results)) : undefined,
```

src/hooks/useConcepts.tsx
```diff
@@ -39,7 +39,10 @@ export function useConcepts(references: Set<string>): {
 
   const results = useMemo(
     () => ({
-      concepts: data ? [].concat(data?.map((res) => res?.data?.results).flat()) : undefined,
+      // data?.[0] check is added for tests, as response is undefined in tests
+      // hence the returned concepts are [undefined], which breaks the form-helper.ts
```
samuelmale, line 43: Makes sense @vasharma05 but since we don't expect such issues in production, does it make sense to fix this in the test env instead? Like maybe mocking the `useSWRInfinite` or its wrapper to behave as expected?

src/hooks/useConcepts.tsx
vasharma05: Hi @samuelmale!

`[].concat(data.map((res) => res?.data?.results))` will return [[Resp, Resp, Resp]], but we want the final array to be `[Resp, Resp, Resp]`

`Resp` represents a response set.

src/hooks/useConcepts.tsx
vasharma05: Actually, the present mocks have added the conditions for `openmrsFetch` instead of `useConcepts`. Adding the mock for `useConcepts` will then break those mocks.

vasharma05:
@samuelmale @denniskigen , requesting your review here.
Thanks!

src/hooks/useSystemSetting.ts
```diff
@@ -0,0 +1,36 @@
...
+        isLowContrast: false,
+      });
+    }
+  }, [error]);
+
+  return {
+    systemSetting: data?.data,
+    error: error,
+    isLoading: isLoading,
+    isValueUuid:
```
samuelmale, line 32: Do you need this?

src/hooks/useRestMaxResultsCount.ts
```diff
@@ -0,0 +1,5 @@
+import useSystemSetting from './useSystemSetting';
+
+export default function useRestMaxResultsCount() {
+  return useSystemSetting('webservices.rest.maxResultsDefault');
+}
```
samuelmale, line 5: I don't see a good reason for having an extra (this) hook. Can we just use the `useSystemSetting(setting)` hook directly?

samuelmale (approved):
LGTM! left a few minor comments.

samuelmale:
I will proceed with merging this. Thanks @vasharma05!

PR #261, opened 2024-05-08
(feat) Add validation for patientIdentifier question type fields when identifier type not set

gitcliff:
## Requirements

## Summary
We need some kind validation around the config ie. If I configured the question to be type patient identifier, the patient identifier type should be a required prop.
## Screenshots
](https://www.loom.com/share/f72a1d4532a74563a1e3815f66b3aa79)


src/validators/form-validator.ts
```diff
@@ -6,15 +7,33 @@ export const fieldOutOfBoundErrCode = 'field.outOfBound';
...
+  validate: (field: FormField, value: any, formValues: Record<string, any>) => {
+    const IDENTIFIRE_TYPE_TITLE = 'Error saving patient Identifier';
+    const IDENTIFIER_TYPE_REQUIRED = 'IdentifierType prop  is required for patientIdentifier fields';
     if (field['submission']?.unspecified) {
       return [];
     }
+    if (field.type === 'patientIdentifier') {
+      // Check if identifierType is not provided or empty
+      if (!field.questionOptions?.identifierType) {
+        showToast({
```
kajambiya, line 19: @gitcliff I think at this level you want to return the `addError` function so that this error is handled the same way all the rest are.

samuelmale: @kajambiya you have a valid point except this isn't the normal validation of user data; this is because the "patient identifier type" isn't user defined but rather defined by the form designer. In a scenario where the ID type wasn't configured, we shouldn't bother saving the ID but we should be able to submit the form (using `addError()` would block submission). This ticket is about informing the user that we won't be able to create an ID because of a missing required prop. @gitcliff I think this should happen at an earlier stage (at field initialisation maybe) and we could disable the field.

hadijahkyampeire: @gitcliff I think you should not use a toast for this, either use an inline notification or the snack bar, toast is now reserved for system errors/notifications. Also while viewing the recording, it looked weird to show a snack bar and a toast.

gitcliff: > @kajambiya you make a valid point however, this isn't the normal validation of user data; this is because the "patient identifier type" isn't user defined but rather defined by the form designer. In a scenario where the ID type wasn't configured, we shouldn't bother saving the ID but we should be able to submit the form (using `addError()` would block submission). This ticket is about informing the user that we won't be able to create an ID because of a missing required prop. @gitcliff I think this should happen at an earlier stage (at field initialisation maybe) and we could disable the field.

@samuelmale @kajambiya i have made some changes in the useEffect hook that initializes the form fields to check if the ID type  was not configured then it disables the field and drops a notification using a snack bar . Added a check during the patient identifier preparation to only allow fields of type patientIdentifier and have set the identifierType . This way only those fields will processed to be saved.

gitcliff:
@kajambiya closing off this and moving the work to the form-builder[ here ](https://github.com/openmrs/openmrs-esm-form-builder/pull/264) as advised by @samuelmale

PR #200, opened 2024-04-22
O3-3058: Support for Conditional Required Validation

reagan-meant:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [ ] My work includes tests or is validated by existing tests.

## Summary

Added Support for Conditional Required Validation

```json
            {
              "label": "Was this visit scheduled?",
              "id": "scheduledVisit",
              "questionOptions": {
                "rendering": "select",
                "concept": "a89ff9a6-1350-11df-a1f1-0026b9348838",
                "answers": [
                  {
                    "concept": "a89b6440-1350-11df-a1f1-0026b9348838",
                    "label": "Scheduled visit"
                  },
                  {
                    "concept": "a89ff816-1350-11df-a1f1-0026b9348838",
                    "label": "Unscheduled Visit Early"
                  },
                  {
                    "concept": "a89ff8de-1350-11df-a1f1-0026b9348838",
                    "label": "Unscheduled Visit Late"
                  }
                ]
              },
              "type": "obs",
              "validators": []
            },
            {
              "label": "If Unscheduled, actual scheduled date",
              "id": "actualDate2",
              "type": "obs",
              "required": {
                "type": "conditionalRequired",
                "message": "Patient visit marked as unscheduled. Please provide the scheduled date.",
                "referenceQuestionId": "scheduledVisit",
                "referenceQuestionAnswers": [
                  "a89ff816-1350-11df-a1f1-0026b9348838",
                  "a89ff8de-1350-11df-a1f1-0026b9348838"
                ]
              },
              "questionOptions": {
                "rendering": "number",
                "concept": "dc1942b2-5e50-4adc-949d-ad6c905f054e"
              },
              "validators": [
                {
                  "type": "date",
                  "allowFutureDates": "true"
                }
              ],
              "hide": {
                "hideWhenExpression": "!arrayContains(['a89ff816-1350-11df-a1f1-0026b9348838','a89ff8de-1350-11df-a1f1-0026b9348838'], scheduledVisit)"
              }
            },
            {
[...]

PR #265, opened 2024-05-09
O3-3058: Support for Conditional Required Validation

reagan-meant:
## Requirements

## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/47450907/b52eaf47-dc2e-4979-830e-cd937933af92

## Related Issue

https://openmrs.atlassian.net/browse/O3-3058

### DIFF
--- a/src/components/inputs/radio/radio.scss
+++ b/src/components/inputs/radio/radio.scss
@@ -25,3 +25,12 @@
   font-weight: 400;
   color: colors.$red-60;
 }
+
+.errorMessage {
+  display: block;
+  max-height: 12.5rem;
+  overflow: visible;
+  font-weight: 400;
+  color: colors.$red-60;
+  font-size: 0.800rem;
+}
--- a/src/utils/boolean-utils.ts
+++ b/src/utils/boolean-utils.ts
@@ -25,10 +25,10 @@ export function isUuid(value: string): boolean {
   return uuidRegex.test(value);
 }
 
-export function isFieldRequired(value: FormField, values): boolean {
-  if (typeof value.required !== 'object' ){
-    return  isTrue(value.required);
-  } else if ( typeof value.required === 'object' && value.required.type === 'conditionalRequired' ){
-    return  value.required?.referenceQuestionAnswers.includes(values[value.required?.referenceQuestionId])    ;
+export function isFieldRequired(field: FormField, values): boolean {
+  if (typeof field.required !== 'object' ){
+    return  isTrue(field.required);
+  } else if ( typeof field.required === 'object' && field.required.type === 'conditionalRequired' ){
+    return  field.required?.referenceQuestionAnswers.includes(values[field.required?.referenceQuestionId])    ;
   }
 }
--- a/src/utils/boolean-utils.ts
+++ b/src/utils/boolean-utils.ts
@@ -25,10 +25,10 @@ export function isUuid(value: string): boolean {
   return uuidRegex.test(value);
 }
 
-export function isFieldRequired(value: FormField): boolean {
+export function isFieldRequired(value: FormField, values): boolean {
   if (typeof value.required !== 'object' ){
     return  isTrue(value.required);
   } else if ( typeof value.required === 'object' && value.required.type === 'conditionalRequired' ){
-    return true;
+    return  value.required?.referenceQuestionAnswers.includes(values[value.required?.referenceQuestionId])    ;
   }
 }
--- a/src/utils/boolean-utils.ts
+++ b/src/utils/boolean-utils.ts
@@ -1,4 +1,4 @@
-import type {  FormField } from "../types";
+import type { FormField } from '../types';
 /**
  * Evaluates whether a value is truthy. This should be used when a string value is expected to parsed into a boolean ie.
  * ```bash
@@ -23,12 +23,4 @@ export function isTrue(value: string | boolean): boolean {
 export function isUuid(value: string): boolean {
   const uuidRegex = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
   return uuidRegex.test(value);
-}
-
-export function isFieldRequired(field: FormField, values): boolean {
-  if (typeof field.required !== 'object' ){
-    return  isTrue(field.required);
-  } else if ( typeof field.required === 'object' && field.required.type === 'conditionalRequired' ){
-    return  field.required?.referenceQuestionAnswers.includes(values[field.required?.referenceQuestionId])    ;
-  }
 }
--- a/src/utils/boolean-utils.ts
+++ b/src/utils/boolean-utils.ts
@@ -1,3 +1,4 @@
+import type {  FormField } from "../types";
 /**
  * Evaluates whether a value is truthy. This should be used when a string value is expected to parsed into a boolean ie.
  * ```bash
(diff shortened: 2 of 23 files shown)


src/validators/form-validator.ts
```diff
@@ -3,15 +3,18 @@ import { isTrue } from '../utils/boolean-utils';
...
       return [];
     }
-    if (isTrue(field.required) || isTrue(field.unspecified)) {
-      if (isEmpty(value)) {
-        return addError(fieldRequiredErrCode, 'Field is mandatory');
+    if (isEmpty(value)) {
+      if (  (typeof field.required === 'string' || typeof field.required === 'boolean' && isTrue(field.required)) || isTrue(field.unspecified)) {
+          return addError(fieldRequiredErrCode, 'Field is mandatory');
+      } else if (typeof field.required === 'object' && field.required?.type === "conditionalRequired" && !isEmpty(formValues) &&  field.required?.referenceQuestionAnswers.includes(formValues[field.required?.referenceQuestionId])) {
+            return addError(fieldConditionalRequiredErrCode, field.required.message);
```
samuelmale, line 17: You have formatting issues. I would also simplify the logic above to:

```ts
  if (isEmpty(value) && (field.required || field.unspecified)) {
      if (typeof field.required !== 'object' && (isTrue(field.required) || isTrue(field.unspecified))) {
        return addError(fieldRequiredErrCode, 'Field is mandatory');
      } else if (
        field.required?.type === 'conditionalRequired' &&
        field.required?.referenceQuestionAnswers.includes(formValues[field.required?.referenceQuestionId])
      ) {
        return addError(fieldRequiredErrCode, field.required.message ?? 'Field is mandatory');
      }
    }

```

src/validators/form-validator.ts
```diff
@@ -3,15 +3,18 @@ import { isTrue } from '../utils/boolean-utils';
...
-  validate: (field: FormField, value: any) => {
+  validate: (field: FormField, value: any, formValues: Record<string, any> ) => {
     if (field['submission']?.unspecified) {
       return [];
     }
-    if (isTrue(field.required) || isTrue(field.unspecified)) {
-      if (isEmpty(value)) {
-        return addError(fieldRequiredErrCode, 'Field is mandatory');
+    if (isEmpty(value)) {
+      if (  (typeof field.required === 'string' || typeof field.required === 'boolean' && isTrue(field.required)) || isTrue(field.unspecified)) {
```
samuelmale, line 14: In the current implementation we do something like:

```ts
  if (isTrue(field.required) || isTrue(field.unspecified)) {
      if (isEmpty(value)) {
        return addError(fieldRequiredErrCode, 'Field is mandatory');
      }
    }

```

If a field has an empty value, we check whether it's required or unspecified. @ebambo it's not clear what the role of "unspecified" is. Does it simply let you bypass the mandatory behaviour of a field? Is it a implicit way of marking a field as required? Or both of the above?

src/validators/form-validator.ts
```diff
@@ -3,15 +3,18 @@ import { isTrue } from '../utils/boolean-utils';
 
 export const fieldRequiredErrCode = 'field.required';
 export const fieldOutOfBoundErrCode = 'field.outOfBound';
+export const fieldConditionalRequiredErrCode = 'field.conditionalRequired';
```
samuelmale, line 6: What stops you from using `fieldRequiredErrCode`?

samuelmale:
@reagan-meant can you add a demo reflecting the latest changes? It's also important to give a brief description of what this PR does.

src/validators/form-validator.ts
ebambo: @samuelmale the `unspecified` is a way of bypassing the mandatory behaviour of a filed. This is to provide the user the ability of not filling a particular field if no information is available on the paper form. For some fields you can give clear options when there is missing information and you need to know why the missing information. An alternative to this would be not marking a field as required but, in this case, you won't know if the user did not fill the filed because he just missed the field or there was no information available. With unspecified if a particular field is empty you know that was the user decision not to fill that field. Hope this is clear

ebambo:
@reagan-meant @samuelmale what is the difference between this PR and the one that was already merged [here](https://github.com/openmrs/openmrs-form-engine-lib/pull/200)?

samuelmale:
> @reagan-meant @samuelmale what is the difference between this PR and the one that was already merged [here](https://github.com/openmrs/openmrs-form-engine-lib/pull/200)?

It was later reverted, [see](https://github.com/openmrs/openmrs-form-engine-lib/pull/265#issuecomment-2106925408)

src/components/inputs/radio/radio.scss
```diff
@@ -25,3 +25,12 @@
...
 }
+
+.errorMessage {
+  display: block;
+  max-height: 12.5rem;
+  overflow: visible;
+  font-weight: 400;
+  color: colors.$red-60;
+  font-size: 0.800rem;
+}
```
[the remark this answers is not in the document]
reagan-meant: Attached in the demo. It is the last field related to radio button

reagan-meant:
Uploading Screen Recording 2024-05-13 at 20.42.30.mov…

src/validators/form-validator.ts
samuelmale: So in summary, this property just let's us bypass the mandatory behaviour but not mark a field as mandatory. Eg. The field below shouldn't be required. 

```json
{
  "label": "Test Results",
  "id": "__L42EMuowp",
  "questionOptions": {
    "rendering": "select",
    "concept": "192f67ac-247c-4e1d-966a-4c223e33d38b",
    "answers": [
    ]
  },
  "required": false,
  "unspecified": true,
  "type": "obs"
}

```

Hence we should be doing:

```ts

  if (isTrue(field.required) /**|| isTrue(field.unspecified) **/) {
      if (isEmpty(value)) {
        return addError(fieldRequiredErrCode, 'Field is mandatory');
      }
    }

```

cc: @reagan-meant

src/utils/boolean-utils.ts
```diff
@@ -23,3 +24,11 @@ export function isUuid(value: string): boolean {
   const uuidRegex = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
   return uuidRegex.test(value);
 }
+
+export function isFieldRequired(value: FormField): boolean {
```
samuelmale, line 28:
```suggestion
export function isFieldRequired(field: FormField): boolean {
```

src/utils/boolean-utils.ts
```diff
@@ -23,3 +24,11 @@ export function isUuid(value: string): boolean {
...
+  if (typeof value.required !== 'object' ){
+    return  isTrue(value.required);
+  } else if ( typeof value.required === 'object' && value.required.type === 'conditionalRequired' ){
+    return true;
```
samuelmale, line 32: You have formatting issues in the second "if" clause. 

Two questions:
1. What happens if the underlying condition (the condition that determines the mandatory behaviour) is not met or resolves to false?
2. What happens if the second "if" resolves to false (I know it's less likely but we should return false or better just return the expression)?

reagan-meant:
https://github.com/openmrs/openmrs-form-engine-lib/assets/47450907/2a10087e-7e89-47ee-8dcb-6647601c3de0

samuelmale:
The attached GIF demonstrates how this works. Note that the labels are no longer being highlighted. Any objections to this? cc: @denniskigen @ibacher

samuelmale:
Looks like the GIF is corrupted. I should retire GIF brewery!!

ibacher:
From the GIF, that looks great!

samuelmale (approved):

PR #267, opened 2024-05-10
(feat) O3-3165: Add modal confirmation modal when deleting repeated question

NethmiRodrigo:
## Requirements

## Summary

This PR implements the functionality of triggering a confirmation modal whenever a user tries to delete a obsGroup repeated rendering question.
The `form-engine-component` has been updated to accept a new prop for the on click handler when deleting a question. Based on the prop, the component either calls the function passed in as the prop (which renders the modal) or proceeds with the deletion of the question (without showing a modal).

## Screenshots

Screenshot 2024-05-24 at 6 35 13 PM

## Related Issue

[Jira Ticket](https://openmrs.atlassian.net/browse/O3-3165)
[Related PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1840)

## Other

While this PR covers the intended functionality, it does bring forth some issues,
1. The carbon version needed to be changed as it caused issues in the modal styling (namely the placement of the close icon, refer screenshot below). 
Screenshot 2024-05-14 at 8 06 47 PM


src/components/repeat/repeat.component.tsx
```diff
@@ -25,6 +26,8 @@ const Repeat: React.FC<FormFieldProps> = ({ question, onChange, handler }) => {
   const [counter, setCounter] = useState(0);
   const [rows, setRows] = useState([]);
   const [fieldComponent, setFieldComponent] = useState(null);
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+  const [selectedQuestion, setSelectedQuestion] = useState(null);
```
samuelmale, line 30: `questionToDelete` reads better

samuelmale:
The "delete" button of the modal should be coloured red.

src/components/repeat/repeat.component.tsx
NethmiRodrigo: Yep it sounds much better. Updated the names accordingly :D

NethmiRodrigo:
@samuelmale should we also show a snackbar/toast message after the user has pressed the confirm button on the modal?

src/form-engine.component.tsx
```diff
@@ -69,6 +76,10 @@ export interface FormSubmissionHandler {
   validate: (values) => boolean;
 }
 
+const functionStore = createGlobalStore<HandleConfirmDeletionFunctionStore | null>('functionStore', {
```
ibacher, line 79: I think rather than using a global store for this, it makes more sense to use a React context. Context's are local to a React tree, so we have fewer issues around naming conflicts, etc. Global stores are for things we need to be shared across different React trees, but here I don't see a reason for it.

NethmiRodrigo: Thanks, Ian. I've updated to using context instead, as suggested. 
Could you please review the context part? I hope I've implemented it right.

ibacher: LGTM. I might rename the context `ExternalFunctionContext` or something so it's reusable for other similar functions (since I imagine we'll need something like this for other uses). Nice job!

ibacher (approved):

ibacher:
@samuelmale Any objections to merging this?

samuelmale (approved):
LGTM! Nice work @NethmiRodrigo

PR #269, opened 2024-05-13
(fix) make visit information available within expression context

arodidev:
## Requirements

## Summary

This PR aims to provide visit information (`visitTypeUuid`) to the expression context to be used for expression evaluation.


src/components/encounter/encounter-form.component.tsx
```diff
@@ -42,7 +42,7 @@ interface EncounterFormProps {
   formSessionDate: Date;
   provider: string;
   location: SessionLocation;
-  visit?: Visit;
+  visit: Visit;
```
CynthiaKamau, line 45: Are we now making the visit mandatory?

src/form-engine.test.tsx
```diff
@@ -598,14 +598,6 @@ describe('Form engine component', () => {
...
-    render(
-      <FormEngine
-        formJson={formJson}
-        formUUID={formUUID}
-        patientUUID={patientUUID}
-        formSessionIntent={intent}
-        visit={visit}
-      />,
-    );
+    render(<FormEngine formJson={formJson} formUUID={formUUID} patientUUID={patientUUID} formSessionIntent={intent} />);
```
CynthiaKamau, line 601: What happens if its RDE? How do you know the visit  to use if the visit uuid or visit is not passed as a prop?

src/components/encounter/encounter-form.component.tsx
arodidev: Yes, since it is now not being passed as a prop but rather being fetched within the root component, it wouldn't make sense to have it as optional anymore yet it is always available.

src/form-engine.test.tsx
arodidev: I am not entirely sure what RDE is, please explain. For the second part, [the vistit is being fetched using the patientUuid](https://o3-docs.openmrs.org/docs/framework-api-reference#usevisit:~:text=%E2%96%B8%20useVisit(patientUuid)%3A%20VisitReturnType).

samuelmale:
@arodidev does it makes sense to consider investigating why we end up with an undefined visit assuming the form is being launched from the workspace? I'm just worried of the potential performance implications unless the underlying hook is optimised. cc: @ibacher

arodidev:
> why we end up with an undefined visit assuming the form is being launched from the workspace? 

@samuelmale This doesn't seem to be the case from my side.

Screenshot 2024-05-15 at 21 49 58

Screenshot 2024-05-15 at 21 51 38


Could you show how to replicate your findings above?

samuelmale:
@arodidev I didn't understand why it is necessary for the form-engine to load the visit by itself; what are the issues with the current setup?

arodidev:
> @arodidev I didn't understand why it is necessary for the form-engine to load the visit by itself; what are the issues with the current setup?

Well, considering we are getting the visit from the `useVisit` hook, deprecating the other one was just a means of removing redundancy. Should we maintain both?

ibacher:
Ideally, I think the form engine wouldn't be calling `useVisit()`, but have the visit passed in. Adding it here is essentially adding a redundant API call. Both visit and encounter should be determined from the context the form is being entered in, rather than by the form engine itself, so it's reasonable to expect visit and (optionally) encounter to be passed as props.

src/utils/expression-runner.ts
```diff
@@ -60,6 +64,7 @@ export function evaluateExpression(
     sex,
     age,
     HD,
+    visitTypeUuid,
```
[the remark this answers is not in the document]
arodidev: Done @samuelmale

PR #272, opened 2024-05-15
(fix) Fix rules of hooks violations

denniskigen:
## Requirements

## Summary

Fixes two violations of the [rules of hooks](https://react.dev/reference/rules/rules-of-hooks) as flagged by ESLint via the react-hooks rules of hooks rule.

## Screenshots

CleanShot 2024-05-16 at 2  07 52@2x


src/post-submission-actions/program-enrollment-action.ts
```diff
@@ -1,14 +1,13 @@
...
-import dayjs from 'dayjs';
 import { type PostSubmissionAction, type ProgramEnrollmentPayload } from '../types';
-import { useTranslation } from 'react-i18next';
 
 export const ProgramEnrollmentSubmissionAction: PostSubmissionAction = {
   applyAction: async function ({ patient, encounters, sessionMode }, config) {
     const encounter = encounters[0];
     const encounterLocation = encounter.location['uuid'];
-    const { t } = useTranslation();
+    const t = window.i18next.t;
```
denniskigen, line 10: @ibacher is there a better approach for getting the `t` for use outside of React components? I don't remember. Also, I've not tested whether this works as expected.

ibacher: If you add i18next as a peer dep, you should just be able to do:

```
import { t } from ‘i18next’;
```

Which is cleaner. 

In normal components, we set the default namespace via a I18nextProvider. The raw `t()` function won’t pull from that context, so we’ll need to somehow set the namespace here for things to work as expected; otherwise it’ll always return a null object.

denniskigen: @vasharma05 how can I set the namespace here as Ian suggested?

src/post-submission-actions/program-enrollment-action.ts
```diff
@@ -40,7 +39,11 @@ export const ProgramEnrollmentSubmissionAction: PostSubmissionAction = {
               title: t('enrollmentFailed', 'Enrollment failed'),
               kind: 'error',
               critical: false,
-              description: t('cannotEnrollPatientToProgram', 'This patient is already enrolled in the selected program'),            });
+              description: t(
+                'cannotEnrollPatientToProgram',
+                'This patient is already enrolled in the selected program',
+              ),
```
vasharma05, line 45:
```suggestion
              description: translateFrom(
              	 '@openmrs/esm-form-engine-lib',
                'cannotEnrollPatientToProgram',
                'This patient is already enrolled in the selected program',
              ),
```

samuelmale: Hhmmm curious whether that would work. We may have to add the translations to the form engine app and instead:

```ts
description: translateFrom(
              	 '@openmrs/esm-form-engine-app',
                'cannotEnrollPatientToProgram',
                'This patient is already enrolled in the selected program',
              )

```

vasharma05: > We may have to add the translations to the form engine app and instead:

Wait, why?

samuelmale: Because it's not part of the import-map so at startup, the app-shell won't pick the resource bundles

vasharma05: @ibacher , we are loading the translations for `@openmrs/form-engine-lib` dynamically as was implemented in this initialization PR for i18n. (specifically [this file](https://github.com/openmrs/openmrs-form-engine-lib/blob/03bb84e9f54a613da467d6f78f0feed68abea440/src/setupI18n.ts#L4))

vasharma05: CC: @samuelmale

vasharma05: FYI, `translateFrom` is basically using window.i18n 

Refer: https://github.com/openmrs/openmrs-esm-core/blob/35438c97f9a54cee71379a7760b050e71383678a/packages/framework/esm-translations/src/index.ts#L40

vasharma05: In short, if we have the translation present in the `{lang}.json` file, then whenever a form is loaded, the {lang}.json is loaded dynamically and hence all the translations in the form-engine-lib is present in the window.i18n, under the namespace `@openmrs/form-engine-lib`
Thanks!

samuelmale: Yes you're right @vasharma05. The i18n setup in the form-engine reconciles with the global one so theoretically it should work.

ibacher: That's right... I forgot about that.

ibacher: @vasharma05 Well, yes, but `translateFrom()` is also basically equivalent to:

```ts
import { i18n: { t } } from 'i18next';

// ...

t('key', { ns: moduleName });
```

Which would be functionally identical.

samuelmale (approved):

src/post-submission-actions/program-enrollment-action.ts
vasharma05: Yes @ibacher, but when I tried this during the hackathon, it didn't work

The `i18next` was different from `window.i18n`, didn't get time to break down into it.

PR #286, opened 2024-05-22
(feat) Introduce locale/Openmrs datepicker

kajambiya:
## Requirements

## Summary
This PR introduces the `OpenmrsDatepicker` component from openmcore/esm-styleguide. Since the Ethiopian team already uses this component in patient registration, there's no need for them to register a different datepicker in Form Engine.

## Screenshots

https://github.com/openmrs/openmrs-form-engine-lib/assets/12844964/017df011-ca0e-4d4e-967e-27a304687c51

### DIFF
--- a/src/components/inputs/date/date.component.tsx
+++ b/src/components/inputs/date/date.component.tsx
@@ -137,6 +137,8 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
                 invalid={errors.length > 0}
                 invalidText={errors[0]?.message}
                 value={field.value}
+                disabled={question.isDisabled}
+                readonly={isTrue(question.readonly)}
                 carbonOptions={{
                   placeholder: placeholder,
                   warn: warnings[0]?.message,
@@ -144,8 +146,6 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
                   className: styles.boldedLabel,
                   datePickerType: datePickerType,
                 }}
-                {...(typeof question.readonly === 'boolean' && { readonly: question.readonly })}
-                {...(typeof question.disabled === 'boolean' && { disabled: question.disabled })}
               />
             </Layer>
           </div>
--- a/src/components/inputs/date/date.component.tsx
+++ b/src/components/inputs/date/date.component.tsx
@@ -2,16 +2,17 @@ import React, { useEffect, useMemo, useState } from 'react';
 import dayjs from 'dayjs';
 import { useTranslation } from 'react-i18next';
 import { useField } from 'formik';
-import { DatePicker, DatePickerInput, Layer, TimePicker } from '@carbon/react';
+import { Layer, TimePicker } from '@carbon/react';
 import { type FormFieldProps } from '../../../types';
 import { isTrue } from '../../../utils/boolean-utils';
 import { isInlineView } from '../../../utils/form-helper';
 import { isEmpty } from '../../../validators/form-validator';
 import { FormContext } from '../../../form-context';
-import { useFieldValidationResults } from '../../../hooks/useFieldValidationResults';
 import FieldValueView from '../../value/view/field-value-view.component';
 import RequiredFieldLabel from '../../required-field-label/required-field-label.component';
 import styles from './date.scss';
+import { useFieldValidationResults } from '../../../hooks/useFieldValidationResults';
+import { OpenmrsDatePicker } from '@openmrs/esm-framework';
 
 const locale = window.i18next.language == 'en' ? 'en-GB' : window.i18next.language;
 const dateFormatter = new Intl.DateTimeFormat(locale);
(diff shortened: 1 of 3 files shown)


arodidev:
Thanks @kajambiya, mind resolving the conficts and build failure?

src/form-engine.test.tsx
```diff
@@ -649,7 +649,7 @@ describe('Form engine component', () => {
       });
     });
 
-    it('should only show question when age is under 5', async () => {
+    fit('should only show question when age is under 5', async () => {
```
samuelmale: Ditto

package.json
```diff
@@ -52,7 +52,7 @@
   },
   "devDependencies": {
     "@carbon/react": "^1.37.0",
-    "@openmrs/esm-framework": "next",
+    "@openmrs/esm-framework": "^5.6.1-pre.1880",
```
samuelmale, line 55: Can you revert this change?

package.json
```diff
@@ -86,7 +86,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "jest-when": "^3.6.0",
     "lint-staged": "^15.2.2",
-    "openmrs": "next",
+    "openmrs": "^5.6.1-pre.1880",
```
samuelmale, line 89: Ditto

src/components/inputs/date/date.component.tsx
```diff
@@ -121,36 +123,30 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
...
+                value={field.value}
+                carbonOptions={{
+                  placeholder: placeholder,
+                  warn: warnings[0]?.message,
+                  warnText: warnings[0]?.message,
+                  className: styles.boldedLabel,
+                  datePickerType: datePickerType,
+                }}
+                {...(typeof question.readonly === 'boolean' && { readonly: question.readonly })}
+                {...(typeof question.disabled === 'boolean' && { disabled: question.disabled })}
```
samuelmale, line 148: Would this work?

```suggestion
                disabled={question.isDisabled}
                readonly={question. readonly}
```

kajambiya: Not really
In RFE, the type of **question.readonly** is `string | boolean` and that of **question.disabled** is `boolean | Omit<DisableProps, "isDisabled">` whereas in core **readonly** and **disabled** are both of type `boolean.`
The solution was either to allow the `OpenmrsDatepicker props` in core accept the same exact types or handle them at FE level. 
I tried to do it in core but according to [this](https://github.com/openmrs/openmrs-esm-core/pull/1007#discussion_r1612280735) comment from @ibacher, this should be handled in FE and core only receives boolean types.

When I went through the code, I realized that for `readonly` for example, it accepts a string (expression) but after the expression is evaluated, the value is changed to a boolean. Given this background, it made sense to only pass these 2 if the data type is boolean so that the types in core stay intact.

package.json
kajambiya: Screenshot 2024-06-06 at 13 07 11
When I change these back to next, the CI fails(see screenshot)

src/components/inputs/date/date.component.tsx
samuelmale: > In RFE, the type of question.readonly is string | boolean and that of question.disabled is boolean | Omit<DisableProps, "isDisabled"> whereas in core readonly and disabled are both of type boolean.

For the disabled property you're supposed to use `question.isDisabled` which is type boolean.

As for `readonly` and other related props we should update the typing to simply `boolean` because the [schema transformer parses the strings to booleans](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/transformers/default-schema-transformer.ts#L79).

kajambiya: Just checked the schema `parseBooleanTokenIfPresent` function. 
From my understanding, it assumes that the the string is either "true" or "false"
However, we have cases where it is a totally different expression (I believe this wasn't a smart implementation) and I've seen it used in PMTCT forms.
Here is [how FE handles it](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/components/encounter/encounter-form.component.tsx#L309C11-L322C12)
We need to agree if this is no longer desirable so that the forms where it is being used can also be adjusted accordingly.

samuelmale: You make a good point but at this point it's still fair to assume that the `readonly` prop has already been processed to a boolean. As you can see in the snippet below the expression is stored in `field['readonlyExpression']`. (I think to be more consistent we should use the `readonly` vs `isReadonly` pattern)

```ts
if (typeof field.readonly == 'string' && field.readonly?.split(' ')?.length > 1) {
            // needed to store the expression for further evaluations
            field['readonlyExpression'] = field.readonly;
            field.readonly = evaluateExpression(

```

To make the TS compiler happy, use the `isTrue(string | boolean)` function:

```ts
readonly={isTrue(question.readonly)}
```

kajambiya: This makes sense.

Done.

samuelmale (approved):
LGTM

PR #305, opened 2024-06-05
(test) Adding test coverage for encounter location handler

gitcliff:
## Requirements

## Summary
Adding test coverage for the encounter location handler for handlingFieldSubmission, getting initial and previous values plus getDisplay value


src/submission-handlers/encounterLocationHandler.test.ts
```diff
@@ -0,0 +1,260 @@
...
+      const field: FormField = {
+        label: 'Encounter Location',
+        type: 'encounterLocation',
+        required: false,
+        id: 'encounterLocation',
+        questionOptions: {
+          rendering: 'ui-select-extended',
+        },
+        validators: [],
+      };
```
pirupius, line 228: @samuelmale is this a pattern that will be maintained? The field definition seems to be repeated throughout the file but it's the same content. Or is it this way so each instance in the test is independent?

samuelmale: I think for this scenario it should be possible to reuse a single field object across the suite. We can then add a cleanup or teardown callback that [ensures that the side effects from the previous tests are cleaned up](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/components/inputs/select/dropdown.test.tsx#L78-L81)

gitcliff: thanks @pirupius @samuelmale ..
i have added a tear down

pirupius (approved):

PR #312, opened 2024-06-07
(feat) O3-2491: Add support for a time picker

NethmiRodrigo:
## Requirements

## Summary

This PR aims to support the rendering of a time picker alongside the date picker based on the prop `datePickerFormat`. If the prop is null, only the date picker will be displayed.

## Screenshots

Screenshot 2024-06-07 at 5 24 30 PM

**Submitting values**

https://github.com/openmrs/openmrs-form-engine-lib/assets/34070216/6d9f844b-3b3b-4574-ac9e-d3af8fa29cc5

**Editing time**

https://github.com/openmrs/openmrs-form-engine-lib/assets/34070216/f17c41cb-7523-41d2-baab-3a0ce9030776

## Related Issue

[JIRA ticket](https://issues.openmrs.org/browse/O3-2491)

## Other

[PR to update prop name](https://github.com/openmrs/openmrs-esm-form-builder/pull/290)

### DIFF
--- a/src/transformers/default-schema-transformer.ts
+++ b/src/transformers/default-schema-transformer.ts
@@ -151,7 +151,7 @@ function transformByRendering(question: FormField) {
       question.datePickerFormat = question.datePickerFormat ?? 'calendar';
       break;
     case 'datetime':
-      question.datePickerFormat = 'both';
+      question.datePickerFormat = question.datePickerFormat ?? 'both';
       break;
   }
   return question;
--- a/src/transformers/default-schema-transformer.ts
+++ b/src/transformers/default-schema-transformer.ts
@@ -149,6 +149,10 @@ function transformByRendering(question: FormField) {
       break;
     case 'date':
       question.datePickerFormat = question.datePickerFormat ?? 'calendar';
+      break;
+    case 'datetime':
+      question.datePickerFormat = 'both';
+      break;
   }
   return question;
 }
--- a/src/transformers/default-schema-transformer.ts
+++ b/src/transformers/default-schema-transformer.ts
@@ -147,6 +147,8 @@ function transformByRendering(question: FormField) {
     case 'group':
       handleLabOrders(question);
       break;
+    case 'date':
+      question.datePickerFormat = question.datePickerFormat ?? 'calendar';
   }
   return question;
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -111,6 +111,7 @@ export interface FormField {
   label: string;
   type: string;
   questionOptions: FormQuestionOptions;
+  datePickerFormat?: 'both' | 'calendar' | 'timer';
   id: string;
   groupId?: string;
   questions?: Array<FormField>;
@@ -262,6 +263,7 @@ export type RenderType =
   | 'content-switcher'
   | 'date'
   | 'datetime'
+  | 'time'
   | 'encounter-location'
   | 'encounter-provider'
   | 'encounter-role'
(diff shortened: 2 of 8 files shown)


src/transformers/default-schema-transformer.ts
```diff
@@ -147,6 +147,8 @@ function transformByRendering(question: FormField) {
     case 'group':
       handleLabOrders(question);
       break;
+    case 'date':
+      question.datePickerFormat = question.datePickerFormat ?? 'calendar';
```
samuelmale, line 151: For backwards compatibility, do you want to add a case for "datetime" where we set `datePickerFormat` to "both" assuming "both" renders both the calendar and time picker?

src/types.ts
```diff
@@ -261,6 +262,7 @@ export type RenderType =
   | 'content-switcher'
   | 'date'
   | 'datetime'
+  | 'time'
```
samuelmale, line 265: Here you're introducing a new rendering that is supported through the "date" rendering. With the introduction of the `datePickerFormat` prop, we only need one rendering "date". Meaning we should consider removing or deprecating the "datetime" rendering.

samuelmale (commented):
For cases where `datePickerFormat` is set to "timer", what kind of value will be submitted to the backend? @NethmiRodrigo Can you share a sample payload of an obs?

src/types.ts
NethmiRodrigo: Adding `time` was a mistake so I removed it. Leaving `datetime` as it is for now as it has being referenced in a couple of places.

NethmiRodrigo:
> For cases where `datePickerFormat` is set to "timer", what kind of value will be submitted to the backend? @NethmiRodrigo Can you share a sample payload of an obs?

`{
    "value": "2024-05-27",
    "concept": "81a50873-81c8-4c2e-8892-b2cd635b704c",
    "formFieldNamespace": "rfe-forms",
    "formFieldPath": "rfe-forms-__tyxtEosu2"
}`

Seems like only the date is going through at the moment, fixing it now.

src/types.ts
samuelmale: I think we need to update existing references to ensure that we have a single source of truth. Eg. We should use `datePickerFormat` [here](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/submission-handlers/obsHandler.ts#L163-L165) to determine the date-picker type.

samuelmale:
@NethmiRodrigo can you attach a video or GIF demonstrating a scenario where the date-picker type is set to "timer" in enter vs edit mode?

NethmiRodrigo:
> @NethmiRodrigo can you attach a video or GIF demonstrating a scenario where the date-picker type is set to "timer" in enter vs edit mode?

@samuelmale updated the description with a screen recording of submitting and editing

samuelmale (approved):
LGTM, thanks @NethmiRodrigo!

denniskigen:
Is the consensus that we're keeping the `datetime` rendering type around for compatibility reasons?

samuelmale:
It's fair to treat `datetime` as a deprecated rendering type. But there is backwards compatibility [see](https://github.com/openmrs/openmrs-form-engine-lib/blob/9f2f4ecb0e0b1844f31ae09b868a7782c536444d/src/transformers/default-schema-transformer.ts#L153-L154).

samuelmale:
This is related to this https://github.com/openmrs/openmrs-esm-core/pull/1041. The `labelText` prop was renamed to `label`

ibacher:
I've restored the `labelText` attribute and deprecated the `label` one.

has9h:
Hi @denniskigen @NethmiRodrigo @samuelmale sorry for tagging you all, but I'm not sure what the corresponding PR for this is in the form-builder repo; we will need this into the form-builder to render the time picker, correct?

PR #316, opened 2024-06-10
O3-2252 Calculated values shouldn't be overwritten by the Encounter's…

CynthiaKamau:
… values

## Requirements

## Summary

This PR addresses the issue of null/undefined values on calculated fields in edit mode .
Calculated value is persisted when :
-There is a default value
-The encounter value is undefined
-The encounter value and calculated value are the same, if different

## Screenshots

https://www.loom.com/share/866edea9d4c34b91ab4ef3ef93eba49c?sid=7ffa7c0d-58cd-42e2-a30f-64d4448462d7

### DIFF
--- a/src/hooks/useInitialValues.ts
+++ b/src/hooks/useInitialValues.ts
@@ -23,7 +23,7 @@ export function useInitialValues(
     'encounterDatetime',
     'encounterLocation',
     'patientIdentifier',
-    'encounterRole'
+    'encounterRole',
   ];
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/hooks/useInitialValues.ts
```diff
@@ -77,10 +77,39 @@ export function useInitialValues(
...
+                // Assuming setInitialValues is a function that updates initialValues
+                if (
+                  res === existingVal ||
+                  (res && !existingVal) ||
+                  (res && field.questionOptions.defaultValue)
+                ) {
+                  setInitialValues((prevValues) => ({
+                    ...prevValues,
+                    [field.id]: res,
+                  }));
```
samuelmale, line 102: I would use a similar approach used [here](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/hooks/useInitialValues.ts#L113-L123) for handling async operations or implement a better approach. If the calculated value is empty, we should fallback to picking the value from the encounter.

src/hooks/useInitialValues.ts
```diff
@@ -77,10 +77,39 @@ export function useInitialValues(
...
+                }
+              })
+              .catch((err) => {
+                initialValues[field.id] = err?.message;
```
samuelmale, line 106: Why would you initialise the field with an error message?

samuelmale (commented):
Thanks for working on this @CynthiaKamau, two minor concerns:

- The ticket ID referenced in the PR title doesn't match the one used to name the branch. 
- Can we have some test coverage around this logic?

src/hooks/useInitialValues.ts
CynthiaKamau: Since the evaluation is asynchronous , the returned value will always be a promise

PR #318, opened 2024-06-11
(fix) O3-2252: Calculated values shouldn't be overwritten by the Encounter's values

CynthiaKamau:
… values

## Requirements

## Summary

This PR resolves the issue of null or undefined initial values for calculated fields when entering edit mode. Previously, the assumption was that all calculated fields would be retrieved through asynchronous functions from the expression runner. This assumption led to problems when synchronous functions were used, resulting in null or undefined initial values.

## Screenshots

https://www.loom.com/share/866edea9d4c34b91ab4ef3ef93eba49c?sid=7ffa7c0d-58cd-42e2-a30f-64d4448462d7

### DIFF
--- a/src/hooks/useInitialValues.test.ts
+++ b/src/hooks/useInitialValues.test.ts
@@ -5,6 +5,7 @@ import testEncounter from '__mocks__/use-initial-values/encounter.mock.json';
 import testPatient from '__mocks__/use-initial-values/patient.mock.json';
 import { ObsSubmissionHandler } from '../submission-handlers/obsHandler';
 import { TestOrderSubmissionHandler } from '../submission-handlers/testOrderHandler';
+import { CommonExpressionHelpers } from 'src/utils/common-expression-helpers';
 
 const obsGroupMembers: Array<FormField> = [
   {
@@ -151,7 +152,7 @@ jest.mock('../utils/expression-runner', () => {
 describe('useInitialValues', () => {
   const encounterDate = new Date();
 
-  it('should return empty meaningful defaults in "enter" mode', async () => {
+ it('should return empty meaningful defaults in "enter" mode', async () => {
     let hook = null;
 
     await act(async () => {
@@ -171,7 +172,7 @@ describe('useInitialValues', () => {
             setEncounterProvider: jest.fn,
             setEncounterLocation: jest.fn,
             encounterRole: '',
-            setEncounterRole: jest.fn
+            setEncounterRole: jest.fn,
           },
           formFieldHandlers,
         ),
@@ -210,7 +211,7 @@ describe('useInitialValues', () => {
             setEncounterProvider: jest.fn,
             setEncounterLocation: jest.fn,
             encounterRole: '',
-            setEncounterRole: jest.fn
+            setEncounterRole: jest.fn,
           },
           formFieldHandlers,
         ),
@@ -273,7 +274,7 @@ describe('useInitialValues', () => {
             setEncounterProvider: jest.fn,
             setEncounterLocation: jest.fn,
             encounterRole: '',
-            setEncounterRole: jest.fn
+            setEncounterRole: jest.fn,
           },
           formFieldHandlers,
         ),
@@ -314,7 +315,7 @@ describe('useInitialValues', () => {
             setEncounterProvider: jest.fn,
             setEncounterLocation: jest.fn,
             encounterRole: '8cb3a399-d18b-4b62-aefb-5a0f948a3809',
-            setEncounterRole: jest.fn
+            setEncounterRole: jest.fn,
           },
           formFieldHandlers,
         ),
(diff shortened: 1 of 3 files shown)


src/hooks/useInitialValues.ts
```diff
@@ -100,8 +119,7 @@ export function useInitialValues(
       );
       formFields.push(...flattenedFields);
       setIsEncounterBindingComplete(true);
-      // TODO: Address behaviour in edit mode; see: https://issues.openmrs.org/browse/O3-2252
-      setAsyncInitValues({});
+      setAsyncInitValues({ ...(asyncInitValues ?? {}), ...tempAsyncValues });
```
samuelmale, line 122: Can `asyncInitValues` be null or undefined?

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +378,63 @@ describe('useInitialValues', () => {
...
+    expect(isBindingComplete).toBe(true);
+
+    const weightQuestionConcept = calculateExpressionFields.find((field) => field.id === 'weight').questionOptions
+      .concept;
+    const heightQuestionConcept = calculateExpressionFields.find((field) => field.id === 'height').questionOptions
+      .concept;
+
+    const heightVal = encounter.obs.find((obs) => obs.concept.uuid === heightQuestionConcept).value;
+    const weightVal = encounter.obs.find((obs) => obs.concept.uuid === weightQuestionConcept).value;
+    const calculatedBmi = helper.calcBMI(heightVal, weightVal);
```
samuelmale, line 434: Seems like you're manually evaluating the expressions and initial values here rendering the hook useless. You should be doing something like:

```ts
// verify BMI
expect(initialValues['bmi']).toBe(expectedValue);

// etc..
```

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +378,63 @@ describe('useInitialValues', () => {
...
+
+    expect(heightVal).not.toBeNull();
+    expect(weightVal).not.toBeNull();
+    expect(calculatedBmi).not.toBeNull();
```
samuelmale, line 438: Can we assert the expected values?

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +378,63 @@ describe('useInitialValues', () => {
...
+      bmi: '',
+    };
+
+    const helper = new CommonExpressionHelpers(
+      { value: allFields[1], type: 'field' },
+      {},
+      allFields,
+      valuesMap,
+      allFieldsKeys,
+    );
```
samuelmale, line 398: You shouldn't call this directly

samuelmale (commented):
Incase of a null calculated value, can we fallback using the encounter value?

brandones:
> This PR addresses the issue of null/undefined values on calculated fields in edit mode .
Calculated value is persisted when :
-There is a default value
-The encounter value is undefined
-The encounter value and calculated value are the same, if different

@CynthiaKamau Please strive for greater clarity in PR summaries. "This PR addresses the issue of null/undefined values on calculated fields in edit mode" what is the issue in question? It is not obvious what "null/undefined values on calculated fields in edit mode" means or why it would be a problem. And the phrase "The encounter value and calculated value are the same, if different" doesn't make sense.

Thanks for your work on this; I hope this is useful feedback.

brandones:
@CynthiaKamau Are the next steps clear to you? Any blockers?

CynthiaKamau:
> @CynthiaKamau Are the next steps clear to you? Any blockers?



> @CynthiaKamau Are the next steps clear to you? Any blockers?

I will update the description.Thanks for the feedback

CynthiaKamau:
@samuelmale do you still have any reservations on this ?

src/hooks/useInitialValues.ts
```diff
@@ -77,10 +81,28 @@ export function useInitialValues(
...
           );
+
+          if (field.questionOptions.calculate?.calculateExpression) {
+            const expression = field.questionOptions.calculate.calculateExpression;
+            const node: FormNode = { value: field, type: 'field' };
+            const context = {
+              mode: encounterContext.sessionMode,
+              patient: encounterContext.patient,
+            };
+            if (field.questionOptions.calculate.calculateExpression.includes('resolve')) {
```
samuelmale, line 92: Using "resolve(" would more specific. Can you add a comment as to why this is necessary?

```suggestion
            if (field.questionOptions.calculate.calculateExpression.includes('resolve(')) {
```

samuelmale (commented):
Nice work @CynthiaKamau, just a couple of requests:

1. For cases where we have an encounter value but the value from evaluating the calculate expression is null/undefined, can we callback to the encounter value?
2. Can we have a test case scenario asserting the above?
3. Can we have a test case scenario with an async calculate expressions?

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,228 @@ describe('useInitialValues', () => {
     expect(allFormFields.find((field) => field.id === 'testOrder_1')).not.toBeNull();
     expect(allFormFields.find((field) => field.id === 'testOrder_2')).not.toBeNull();
   });
+
+  it('should return syncronous calculated values for calculated fields in "edit" mode', async () => {
```
samuelmale, line 338:
```suggestion
  it('should return synchronous calculated values for calculated fields in "edit" mode', async () => {
```

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,228 @@ describe('useInitialValues', () => {
...
+    let hook = null;
+    let syncronousCalculateExpressionFields: Array<FormField> = [
```
samuelmale, line 340: You have a typo in the word "synchronous" but since the name of the unit test already reflects the fact that we are dealing with synchronous expressions, I would simply use: `formFields`

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,228 @@ describe('useInitialValues', () => {
...
+    const weightVal = initialValues['weight'];
+
+    const calculatedBmi = helper.calcBMI(heightVal, weightVal);
+
+    expect(initialValues['height']).toBe(176);
+    expect(initialValues['weight']).toBe(56);
+    expect(initialValues['bmi']).toBe(calculatedBmi);
+  });
+
+  it('should return asyncronous calculated values for calculated fields in "edit" mode', async () => {
```
samuelmale, line 437:
```suggestion
  it('should return asynchronously calculated values for fields in "edit" mode', async () => {
```

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,228 @@ describe('useInitialValues', () => {
...
+      );
+    });
+    const {
+      current: { initialValues, isBindingComplete },
+    } = hook.result;
+    expect(isBindingComplete).toBe(true);
+    expect(initialValues['latest_mother_hiv_status']).toBe('664AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');
+  });
+
+  it('should return the encounter value if the calculated values are null/endefined for calculated fields in "edit" mode', async () => {
```
samuelmale, line 483:
```suggestion
 it('should fall back to encounter value if the calculated expression result is null or undefined', () => {
```

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,228 @@ describe('useInitialValues', () => {
...
+    let hook = null;
+    let syncronousCalculateExpressionFields: Array<FormField> = [
```
samuelmale, line 485: Typo

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,228 @@ describe('useInitialValues', () => {
...
+      );
+    });
+    const {
+      current: { initialValues, isBindingComplete },
+    } = hook.result;
+    expect(isBindingComplete).toBe(true);
+
+    expect(initialValues['height']).toBe(176);
+    expect(initialValues['weight']).toBe(56);
+    expect(initialValues['bmi']).toBe(2);
```
samuelmale, line 559: How does one validate that this value is picked from the encounter?

src/hooks/useInitialValues.ts
```diff
@@ -35,13 +35,20 @@ export function useInitialValues(
...
               formFieldHandlers[field.type].handleFieldSubmission(field, result, encounterContext);
             }
           } catch (error) {
+            const encounterValue = formFieldHandlers[field.type]?.getInitialValue(
+              encounter,
+              field,
+              formFields,
+              encounterContext,
+            );
+            formFieldHandlers[field.type].handleFieldSubmission(field, encounterValue, encounterContext);
```
samuelmale, line 51: So we are only falling back to the encounter's value when we run into an error?

CynthiaKamau: yes

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,239 @@ describe('useInitialValues', () => {
...
+        rendering: 'fixed-value',
+        concept: 'af7c1fe6-d669-414e-b066-e9733f0de7a8',
+        calculate: {
+          calculateExpression:
+            "resolve(api.getLatestObs(patient.id, '159427AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA', '2549af50-75c8-4aeb-87ca-4bb2cef6c69a'))?.valueCodeableConcept?.coding[0]?.code",
+        },
+      },
+      id: 'latest_mother_hiv_status',
+    };
+    allFormFields.push(fieldWithCalculateExpression);
```
brandones, line 452: This mutates the object at the top level of this test. This will lead to behavior that varies depending on which tests are run, or the order in which they are run. Objects at the top level of a test file should be considered immutable. Use the spread syntax to define new form fields within specific tests.

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,239 @@ describe('useInitialValues', () => {
...
+    const {
+      current: { initialValues, isBindingComplete },
+    } = hook.result;
+    expect(isBindingComplete).toBe(true);
+    expect(initialValues['latest_mother_hiv_status']).toBe('664AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');
+  });
+
+  it('should fall back to encounter value if the calculated expression result is null or undefined', async () => {
+    let hook = null;
+    let formFields: Array<FormField> = [
```
brandones, line 485: This looks very similar to the other `formFields` variable. It would be easier to understand what the difference is if you defined this at the top level of the test file and then used it in each test file where it is needed, building new modified copies using the object spread syntax as necessary.

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,239 @@ describe('useInitialValues', () => {
...
+            calculateExpression: 'calcBMI(heigh,weight)',
+          },
+        },
+        validators: [],
+        questionInfo: 'this calculates BMI using calcBMI function and useFieldValue of weight and height',
+      },
+    ];
+
+    await act(async () => {
+      hook = renderHook(() =>
```
brandones, line 530: We have this block identical or nearly identical many times in this file. Could you please refactor it out into a helper function?

brandones (changes requested):

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,239 @@ describe('useInitialValues', () => {
...
+        required: false,
+        id: 'bmi',
+        questionOptions: {
+          rendering: 'number',
+          defaultValue: 0,
+          concept: '1342AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+          isTransient: true,
+          disallowDecimals: false,
+          calculate: {
+            calculateExpression: 'calcBMI(heigh,weight)',
```
brandones, line 521: Is "height" intentionally mistyped, is that how you are getting it to produce a null value? That's very easy for a reader of this code to miss. Please either add a comment that points this out

```suggestion
            calculateExpression: 'calcBMI(heigh,weight)',  // This expression has an intentional typo
```

or make it more obviously erroneous

```suggestion
            calculateExpression: 'whoops()',
```

brandones: Of course, following the advice of my previous suggestion, to factor out the form field definition and just override the lines that need overriding for each test, will also help make the intention clearer.

src/hooks/useInitialValues.test.ts
```diff
@@ -333,4 +334,239 @@ describe('useInitialValues', () => {
...
+      const formFieldNamespace = field.formFieldNamespace;
+      const formFieldPath = field.formFieldPath;
+            if (formFieldNamespace && formFieldPath.startsWith(formFieldNamespace)) {
+        const valueAfterNamespace = formFieldPath.slice(formFieldNamespace.length + 1);
+        return valueAfterNamespace === "bmi";
+      }
+      
+      return false;
+    });
+    expect(initialValues['bmi']).toBe(bmiField.value);
```
brandones, line 570: It was better before. I might be wrong, but I think the point of @samuelmale 's question was not "you should pro grammatically get the value from the mock," which has resulted in a very unpleasant little block of code, but rather "you should create a situation in which it is obvious that the value cannot have been calculated." Which is to say, enter a new height or weight such that the calculated BMI should be different from the observed BMI field value (which comes from the encounter). It is not clear to me that this test would fail if the value were being calculated from the field values, which it should. As @samuelmale said, it is not clear that the BMI number in the encounter is different than what would be produced by the calculation.

You can leave in the code that extracts the value from the mock, or change it back to how it was, I don't care about that.

brandones (changes requested):

brandones (approved):

PR #326, opened 2024-06-18
(fix) O3-3350: Resolve infinite re-rendering of the dropdown component in the tablet viewport

samuelmale:
## Requirements

## Summary

This PR fixes an issue with the dropdown component re-rendering infinitely on selecting an option in tablet view port. This issue [stems from the "downshift"](https://github.com/downshift-js/downshift/issues/1511) library consumed by Carbon and it seems to be affecting React v18. It was since resolved [in downshift](https://github.com/downshift-js/downshift/pull/1558) `v8.2.4` and [carbon](https://github.com/carbon-design-system/carbon/pull/15454). 

### Out of scope

I've also changed Carbon from a peer dependency to a normal dependency.

## Related Issue

https://issues.openmrs.org/browse/O3-3350

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@carbon/react": ">1.47.0 <1.50.0",
     "ace-builds": "^1.33.2",
     "classnames": "^2.5.1",
     "dayjs": "1.x",
@@ -41,7 +42,6 @@
     "yup": "^1.4.0"
   },
   "peerDependencies": {
-    "@carbon/react": "1.x",
     "@openmrs/esm-framework": "5.x",
     "@openmrs/esm-patient-common-lib": "8.x",
     "dayjs": "1.x",
@@ -52,7 +52,6 @@
     "swr": "2.x"
   },
   "devDependencies": {
-    "@carbon/react": "^1.37.0",
     "@openmrs/esm-framework": "next",
     "@openmrs/esm-patient-common-lib": "next",
     "@swc/cli": "^0.1.65",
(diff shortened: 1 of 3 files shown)


arodidev (approved):
Thanks @samuelmale

denniskigen:
Just tested on dev3 and this fix works as intended! Thanks, @samuelmale.

package.json
```diff
@@ -29,6 +29,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@carbon/react": ">1.47.0 <1.50.0",
```
pirupius, line 32: @samuelmale Is there a specific reason you made this a dependency and not a peerDependency?

I've read the linked issues/pr and the issue with the downshift dependency in carbon which has since been resolved. After pulling the latest changes and linking with the form-builder, i got some errors from carbon because of version conflicts. 


1. On one end adding this range resolved the issue with the downshift library but it also creates either a conflict or multiple versions of `@carbon/react` running within microfrontends that have dependencies out of that range for example esm-form-builder and other esms are currently running on `@carbon/react@npm:1.37.0` which will in turn affect the bundle size. Would this necessitate us updating the frontends that pull the FE directly to at least `1.47.0` to have this fix working?

Screenshot 2024-06-21 at 11 02 10


2. Doesn't having the same range of versions as a peer dependency yield the same result but with the benefit of not duplicating instances of `@carbon/react`?

samuelmale: IIRC, the overwhelming consensus on the dependency management strategy for Carbon in O3 is having it as a normal dependency as opposed to a singleton (peer).

PR #327, opened 2024-06-18
(fix) O3-3337 - Support expressions that don't use whitespace as delimiters

arodidev:
## Requirements

## Summary

This PR introduces a fix for various failing expression evaluations due to improper formatting during form construction.
## Related Issue

https://openmrs.atlassian.net/browse/O3-3337

### DIFF
--- a/src/utils/expression-parser.test.ts
+++ b/src/utils/expression-parser.test.ts
@@ -7,8 +7,14 @@ import {
   linkReferencedFieldValues,
   parseExpression,
   replaceFieldRefWithValuePath,
+  expressionFormatter,
 } from './expression-parser';
 import { testFields } from './expression-runner.test';
+
+const testFieldValues = {
+  patientIdentificationNumber: 'test value 1',
+  linkedToCare: 'test value 2',
+};
 
 describe('Expression parsing', () => {
   it('should split expression 1 into parts correctly', () => {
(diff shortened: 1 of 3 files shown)


src/utils/expression-runner.ts
```diff
@@ -75,7 +80,7 @@ export function evaluateExpression(
     _,
   };
 
-  expression = linkReferencedFieldValues(fields, fieldValues, parts);
+  expression = expressionFormatter(linkReferencedFieldValues(fields, fieldValues, parts), fields, fieldValues);
```
pirupius, line 83: @arodidev based on our conversion a few days ago. Does the splitting remain the same? and the only difference is the expression being resolved regardless if you have spaces or not?

arodidev: @pirupius pius Yes, that is correct. The formatting of the expression remains unaffected, but rather a `formField` namespace is prefixed to the field id.

src/utils/expression-parser.test.ts
```diff
@@ -306,3 +312,36 @@ describe('hasParentheses', () => {
     expect(hasParentheses(expression)).toBe(true);
   });
 });
+
+describe('expression formatter function', () => {
+  it('should replace the field id with fieldValue.field id for all form fields', () => {
```
brandones, line 317:
```suggestion
describe('expression formatter function', () => {
  it('should not care about whether there are spaces around operators', () => {
```

brandones: It's better to be precise about what the purpose of your test is.

brandones (approved):
LGTM. Next time you get suggestions on your PRs please integrate them rather than just giving a thumbs up.

arodidev:
Hi @brandones, the work on this PR had stalled since @pirupius had picked up the issue, albeit using a different approach.

brandones:
Wait so should we revert this?

When things like that happen, please leave comments on PRs so that people know what's going on.

samuelmale:
@brandones I would say we revert because this is a half-backed solution. Yes it does the necessary interpolation (further hydrates the expression by updating any lurking referenced field's value path) but it doesn't update the relationships (dependencies) etc. The ideal solution is to properly decompose the expression into parts and we were planning on [using a third party library](https://github.com/EricSmekens/jsep).

samuelmale:
>updating the field dependants was not an area of concern while working on this issue since it was non-problematic.

I would say it is a secondary concern.

PR #334, opened 2024-06-28
(fix) Update onDateChange function

NethmiRodrigo:
## Requirements

## Summary

The onChange function sends in the date as a single value of type `CalendarDate`. In our local date change handler, we were converting this date to a Javascript date with the line: 
`const refinedDate = date instanceof Date ? new Date(date.setHours(0, 0, 0, 0)) : new Date(date);`

This ideally should cause a typescript error saying that `Date` does not accept type `CalendarDate` as a param.
But, due to the way that we passed in the onChange handler to the `OpenmrsDatePicker` and the way that we had defined the local date handler, typescript didn't pick it up and a Javascript date was created without any error.

How we passed in our function to the `onChange` parameter:
`onChange={ (date) => onDateChange( [date] ) }` 

Local date handler's function definition:
`const onDateChange = ( [date] ) => {}`

This PR updates the `onDateChange` handler to fix the because this type mismatch might cause unintended issues later on.

### DIFF
--- a/src/components/inputs/date/date.component.tsx
+++ b/src/components/inputs/date/date.component.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import React, { useEffect, useMemo, useState } from 'react';
 import dayjs from 'dayjs';
 import { useTranslation } from 'react-i18next';
@@ -34,7 +33,6 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
   const onDateChange = (date: CalendarDate) => {
     const refinedDate = new Date(date.toString());
     refinedDate.setHours(0, 0, 0, 0);
-    console.log(refinedDate);
     setTimeIfPresent(refinedDate, time);
     setFieldValue(question.id, refinedDate);
     onChange(question.id, refinedDate, setErrors, setWarnings);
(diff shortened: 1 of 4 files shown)


src/components/inputs/date/date.component.tsx
```diff
@@ -29,8 +30,9 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
     return false;
   }, [encounterContext.sessionMode, question.readonly, question.inlineRendering, layoutType, workspaceLayout]);
 
-  const onDateChange = ([date]) => {
-    const refinedDate = date instanceof Date ? new Date(date.setHours(0, 0, 0, 0)) : new Date(date);
+  const onDateChange = (date: CalendarDate) => {
+    const refinedDate = new Date(date.toString());
```
samuelmale, line 34:
```suggestion
    const refinedDate = new Date(date.toDate(getLocalTimeZone())));
```

(You may have to add the import)

src/components/inputs/date/date.component.tsx
```diff
@@ -29,8 +30,9 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
...
+    refinedDate.setHours(0, 0, 0, 0);
```
samuelmale, line 35: Do we still need this?

ibacher: I'm nervous about anything that zeroes out the time. I think this is very likely to create issues where the date is something other than what the user selected depending on what timezone they're located in, unless we're strictly making sure to treat this value as a value in the UTC timezone.

NethmiRodrigo: Yeah I think @samuelmale is right, we can probably get rid of this since we have a time picker now

samuelmale (approved):
LGTM

PR #337, opened 2024-07-02
(fix) Reworked logic for gravida calculation

arodidev:
## Requirements

## Summary

This PR corrects the gravida calculation logic within the form engine.

### DIFF
--- a/src/utils/common-expression-helpers.ts
+++ b/src/utils/common-expression-helpers.ts
@@ -259,11 +259,24 @@ export class CommonExpressionHelpers {
     return values;
   };
 
+  /**
+   * Calculates the gravida (total number of pregnancies) based on term pregnancies and abortions/miscarriages.
+   *
+   * @param {number|string} parityTerm - The number of term pregnancies.
+   * @param {number|string} parityAbortion - The number of abortions (including miscarriages).
+   * @returns {number} The total number of pregnancies (gravida).
+   * @throws {Error} If either input is not a valid number.
+   *
+   * @example
+   * const gravida = calcGravida(2, 1);
+   * console.log(gravida); // Output: 3
+   */
+
   calcGravida = (parityTerm, parityAbortion) => {
     const term = parseInt(parityTerm, 10);
     const abortion = parseInt(parityAbortion, 10);
 
-    if (isNaN(term) || isNaN(abortion)) {
+    if (!Number.isInteger(term) || !Number.isInteger(abortion)) {
       throw new Error('Both inputs must be valid numbers.');
     }
 
--- a/src/utils/common-expression-helpers.ts
+++ b/src/utils/common-expression-helpers.ts
@@ -260,17 +260,7 @@ export class CommonExpressionHelpers {
   };
 
   calcGravida = (parityTerm, parityAbortion) => {
-    let gravida = 0;
-    if (parityTerm === parseInt(parityTerm)) {
-      gravida = parityTerm + 1;
-    }
-    if (parityAbortion === parseInt(parityAbortion)) {
-      gravida = parityAbortion + 1;
-    }
-    if (parityAbortion === parseInt(parityAbortion) && parityTerm === parseInt(parityTerm)) {
-      parityAbortion + parityTerm + 1;
-    }
-    return gravida;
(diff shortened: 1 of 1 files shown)


src/utils/common-expression-helpers.ts
```diff
@@ -260,17 +260,7 @@ export class CommonExpressionHelpers {
...
-      gravida = parityTerm + 1;
-    }
-    if (parityAbortion === parseInt(parityAbortion)) {
-      gravida = parityAbortion + 1;
-    }
-    if (parityAbortion === parseInt(parityAbortion) && parityTerm === parseInt(parityTerm)) {
-      parityAbortion + parityTerm + 1;
-    }
-    return gravida;
+    return parseInt(parityTerm) + parseInt(parityAbortion);
```
samuelmale, line 263: So in other words the Gravida is the summation of the `parityTerm` with `parityAbortion`?

arodidev: Affirmative.

src/utils/common-expression-helpers.ts
```diff
@@ -260,17 +260,14 @@ export class CommonExpressionHelpers {
   };
 
   calcGravida = (parityTerm, parityAbortion) => {
```
denniskigen, line 262:
```suggestion
/**
 * Calculates the gravida (total number of pregnancies) based on term pregnancies and abortions/miscarriages.
 *
 * @param {number|string} parityTerm - The number of term pregnancies.
 * @param {number|string} parityAbortion - The number of abortions (including miscarriages).
 * @returns {number} The total number of pregnancies (gravida).
 * @throws {Error} If either input is not a valid number.
 *
 * @example
 * const gravida = calcGravida(2, 1);
 * console.log(gravida); // Output: 3
 */
  calcGravida = (parityTerm, parityAbortion) => {
```

src/utils/common-expression-helpers.ts
```diff
@@ -260,17 +260,14 @@ export class CommonExpressionHelpers {
...
-    }
-    if (parityAbortion === parseInt(parityAbortion)) {
-      gravida = parityAbortion + 1;
-    }
-    if (parityAbortion === parseInt(parityAbortion) && parityTerm === parseInt(parityTerm)) {
-      parityAbortion + parityTerm + 1;
+    const term = parseInt(parityTerm, 10);
+    const abortion = parseInt(parityAbortion, 10);
+
+    if (isNaN(term) || isNaN(abortion)) {
```
denniskigen, line 266:
```suggestion
    if (!Number.isInteger(term) || !Number.isInteger(abortion)) 
```

Perhaps [Number.isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) is a better check here since we're specifically dealing with whole numbers. Unlike `isNaN`, which allows any valid number (including decimals), `Number.isInteger` will reject decimals. Additionally, `isNaN` attempts to coerce the input into a number before checking, which can lead to unintended behavior, while `Number.isInteger` will return false for any non-number types. If we decide to consider decimal inputs in the future, reverting to `isNaN` would be more appropriate.

denniskigen (commented):
Mostly LGTM. I've left a suggestion related to using a different function for the number validation and a JSDoc comment string.

src/utils/common-expression-helpers.ts
arodidev: Thanks @denniskigen, this has been resolved.

CynthiaKamau (approved):

src/utils/common-expression-helpers.ts
denniskigen: Per our conversation from Coffee Break call, you might need to consider the current pregnancy in your calculation @arodidev.

src/utils/common-expression-helpers.ts
denniskigen: Per our conversation from Coffee Break call, you might need to consider the current pregnancy in your calculation @arodidev.

arodidev: @denniskigen I did a bit of consultation, and it seems the convention should be to do a summation of the two.

samuelmale (approved):
I'm naive when it comes to the business aspects but the code LGTM

PR #342, opened 2024-07-04
(fix) Update form fetch request to return the latest form by uuid or name

CynthiaKamau:
## Requirements

## Summary

When fetching a form by name, the current form engine returns a list of forms with that name, including retired forms. Using the o3 forms endpoint, it is possible that the returned result might be a retired form. This PR ensures that the latest (most recent) form is fetched instead of the first item in the array, which could be a retired form.

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -4,7 +4,9 @@ import { type OpenmrsForm, type PatientIdentifier, type PatientProgramPayload } 
 import { isUuid } from '../utils/boolean-utils';
 
 export function saveEncounter(abortController: AbortController, payload, encounterUuid?: string) {
-  const url = encounterUuid ? `${restBaseUrl}/encounter/${encounterUuid}?v=${encounterRepresentation}` : `${restBaseUrl}/encounter?v=${encounterRepresentation}`;
+  const url = encounterUuid
+    ? `${restBaseUrl}/encounter/${encounterUuid}?v=${encounterRepresentation}`
+    : `${restBaseUrl}/encounter?v=${encounterRepresentation}`;
 
   return openmrsFetch(url, {
     headers: {
@@ -104,7 +106,7 @@ export async function fetchOpenMRSForm(nameOrUUID: string): Promise<OpenmrsForm 
     return openmrsFormResponse;
   }
   return openmrsFormResponse.results?.length
-    ? openmrsFormResponse.results[0]
+    ? openmrsFormResponse.results[openmrsFormResponse.results?.length - 1]
     : new Error(`Form with ${nameOrUUID} was not found`);
 }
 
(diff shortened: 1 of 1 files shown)


arodidev (approved):
Thanks @CynthiaKamau

samuelmale:
> Using the o3 forms endpoint, it is possible that the returned result might be a retired form.

Can you also switch to using the o3 endpoint? Currently, we are [using the standard "form" resource endpoint](https://github.com/openmrs/openmrs-form-engine-lib/blob/1820076ed77fbbfdd6dce39f83fd87e1f446874f/src/api/api.ts#L100-L102). It maybe worth checking if it supports a filter for excluding retired resources.

CynthiaKamau:
> > Using the o3 forms endpoint, it is possible that the returned result might be a retired form.
> 
> Can you also switch to using the o3 endpoint? Currently, we are [using the standard "form" resource endpoint](https://github.com/openmrs/openmrs-form-engine-lib/blob/1820076ed77fbbfdd6dce39f83fd87e1f446874f/src/api/api.ts#L100-L102). It maybe worth checking if it supports a filter for excluding retired resources.

The response from the o3 endpoint is not the latest form if you fetch a form by a name, you can try test as well

samuelmale:
@CynthiaKamau by latest you meaning in terms of the published form version?

samuelmale:
@CynthiaKamau hhmm!! Maybe this is happening for snapshot versions (unpublished versions)? @ibacher are you aware of this?

CynthiaKamau:
> @CynthiaKamau hhmm!! Maybe this is happening for snapshot versions (unpublished versions)? @ibacher are you aware of this?

Does dev3 run on a snapshot version? 
Screenshot 2024-07-05 at 10 53 59

samuelmale:
@CynthiaKamau by "snapshot" I was referring to form versions not the module

ibacher:
> The response from the o3 endpoint is not the latest form if you fetch a form by a name, you can try test as well

I would recommend passing form UUIDs rather than form names to the O3 Forms service...

samuelmale:
@ibacher I think some folks find it convenient to manage configs using the form names assuming the modelling supports that (as secondary identifiers). The standard form REST API seems to work correctly unlike the o3 endpoint as flagged by @CynthiaKamau. Is it worth investigating why?

ibacher:
>  think some folks find it convenient to manage configs using the form names assuming the modelling supports that 

Right... I'm just saying maybe don't naively pass it to the o3forms until that bug is fixed. A call to the `forms` endpoint to get the UUID doesn't seem that bad...

samuelmale:
Sweet! I just wanted to confirm that you're aware of the bug and maybe at a later point we will move to using the o3 endpoint. It's inevitable in my pov!

samuelmale (approved):
I don't seem to wrap my head around how using a dynamic index fixes the issue, but I'll give this a pass.

ibacher:
Yeah, it would be better, I think, to use the first non-retired one. Relying on the order like this is not guaranteed to work, though it will in the common case.

PR #349, opened 2024-07-11
(BREAKING) Migrate from Formik to RHF

samuelmale:
## Requirements

- [X] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

This PR introduces a major migration from Formik to React Hook Form, along with the implementation of a new form processing architecture. The changes focus on enhancing the form handling capabilities, improving performance, and providing a more flexible and maintainable codebase. The introduction of form processors abstracts the form engine from the underlying domain objects, allowing for more modular and reusable code. Additionally, the field submission handlers have been refactored to form-field adapters for better clarity and naming consistency. Form validation has also been updated to address specific challenges.

#### Major Changes Made

1. **Migration from Formik to React Hook Form**:
   - Replaced Formik with React Hook Form across the project to leverage its performance benefits and simpler API.
   - Updated all form components to use React Hook Form's hooks and methods for managing form state and validation.

2. **Introduction of Form Processors**:
   - Implemented a new form processing architecture to abstract the form engine from the underlying domain objects (e.g., Encounter, Patient).
   - Form processors provide an abstraction layer where the form engine is agnostic about the underlying domain object. This allows for the implementation details of the domain to be defined in specific processors, such as the `EncounterFormProcessor`. These processors handle tasks such as resolving dependencies, initializing form values, and processing submissions, ensuring a clean separation of concerns and enhancing code modularity.

3. **`EncounterFormProcessor` Implementation**:
   - Defined the `EncounterFormProcessor` to manage the encounter-specific logic, including preparing the form schema, processing submissions, and resolving dependencies.
   - Implemented methods to handle the initialization of form values, submission of encounter data, and saving of related entities such as patient identifiers, patient programs, and attachments.
   - Utilized custom hooks (`useCustomHooks`) to manage the loading state and context updates for encounters.

4. **Refactoring of Field Submission Handlers to Form-Field Adapters**:
   - Renamed field submission handlers to form-field adapters to better reflect their purpose and functionality.
   - Added a `tearDown` function to the underlying interface of form-field adapters to handle any necessary cleanup.

5. **Form Validation Management**:
   - Initially planned to use Zod's integration with React Hook Form for validation but encountered challenges in managing errors vs. warnings with different severity levels and handling expression-based validators.
   - Decided to manage form validation manually to better control the validation logic and handle complex validation scenarios for the time being.

## Screenshots

TBD

## Related Issue

https://openmrs.atlassian.net/browse/O3-2906

#### Form Engine Migration Checklist

**Pending tasks**
- [ ] Handle tooltips (https://github.com/openmrs/openmrs-form-engine-lib/pull/351)
- [x] Flesh out other value adapters
    - [x] Obs Date
    - [x] Patient IDs
    - [x] Obs Comment
    - [x] Enc Datetime
    - [x] Enc Roles
    - [x] Enc Provider
    - [x] Enc Location
    - [x] Programs
    - [x] Orders
- [x] Flesh out other input components
    - [x] File & Camera
    - [x] Unspecified 
    - [x] Extension Widget
    - [x] Ui-select-extended
    - [x] Fixed Value
    - [x] Markdown
    - [x] Repeating
    - [x] Group
    - [x] Workspace Launcher
    - [x] Toggle
    - [x] Content switcher
    - [x] Text Area
    - [x] Number
    - [x] Date
    - [x] Radio
    - [x] Multi Select
    - [x] Dropdown
    - [x] Text
- [ ] Implement sidebar
- [ ] Implement patient banner
- [x] Ensure skip logic works as expected
- [ ] Resolve TODOs
- [x] Early QA
- [ ] Unit and Integration testing
- [ ] Documentation
- [ ] QA

**Unmasked issues/bugs in the Migration course**
- [x] Previous values aren’t correctly initialised (tested with date comp, enc-datetime)
- [ ] Errors and warnings aren't rendered correctly for content-switcher rendering
- [ ] File picker doesn’t support multiple files selection (https://openmrs.atlassian.net/browse/O3-3682)
- [x] Encounter date time isn’t editable, and it also doesn’t seem to update the encounter’s date (https://openmrs.atlassian.net/browse/O3-3646)
- [x] The Obs comment feature wasn’t functional
- [x] Number input doesn’t render error messages when validation fails


package.json
```diff
@@ -30,16 +30,19 @@
...
     "dayjs": "1.x",
     "formik": "^2.4.6",
     "lodash-es": "^4.17.21",
     "react-error-boundary": "^4.0.13",
+    "react-hook-form": "^7.52.0",
     "react-markdown": "^7.1.2",
     "react-waypoint": "^10.3.0",
     "react-webcam": "^7.2.0",
-    "yup": "^1.4.0"
+    "yup": "^1.4.0",
```
samuelmale, line 44: Remove yup and formik

gracepotma:
@ibacher @brandones and @denniskigen - Samuel would love your early impressions on this work to migrate the engine to RHF :)

brandones:
Wow, this is a big change, and deals with a lot of stuff I haven't really worked on. Sorry I'm not really in a good position to review this.

slubwama:
@samuelmale resolve conflicts

src/provider/form-factory-provider.tsx
```diff
@@ -0,0 +1,138 @@
...
+  location,
+  provider,
+  visit,
+  children,
+  formSubmissionProps,
+  setCurrentPage,
+  handleConfirmQuestionDeletion,
+}) => {
+  const rootForm = useRef<FormContextProps>();
+  const subForms = useRef<Record<string, FormContextProps>>({});
```
ibacher, line 65: Why are we using Refs here?

src/provider/form-factory-provider.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  useEffect(() => {
+    if (isSubmitting) {
+      // TODO: find a dynamic way of managing the forms processing order
+      const forms = [rootForm.current, ...getSubForms()];
+      // validate all forms
+      const valid = forms.every((formContext) => validateForm(formContext));
+      if (valid) {
+        Promise.all(
+          forms.map((formContext) => formContext.processor.processSubmission(formContext, new AbortController())),
```
ibacher, line 90: I always wonder about AbortControllers like this. If there's no point at which the controller can be cancelled, its not really worth defining one.

src/provider/form-factory-provider.tsx
```diff
@@ -0,0 +1,138 @@
...
+        visit,
+        location,
+        provider,
+        registerForm,
+        getSubForms,
+        getRootForm,
+        setCurrentPage,
+        handleConfirmQuestionDeletion,
+      }}>
+      {formProcessors.current && children}
```
ibacher, line 127: Since we set `formProcessors()` above, won't this condition always be `true`?

src/components/processor-factory/form-processor-factory.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+    sessionMode,
+    layoutType,
+    location,
+    currentProvider: provider,
+    processor,
+    sessionDate,
+    visit,
+  });
+  const { formFields: rawFormFields, conceptReferences } = useFormFields(formJson);
+  const { concepts: formFieldsConcepts, isLoading: isLoadingConcepts } = useConcepts(conceptReferences);
```
ibacher, line 49: `isLoading` here seems to be unused?

ibacher: I'd also add that this was one of the sort of requests I'd hoped we could avoid with the o3forms endpoint, but if that doesn't work for some reason, let me know.

src/components/processor-factory/form-processor-factory.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const formFieldValidators = useFormFieldValidators(rawFormFields);
+  const { isLoading: isLoadingCustomDeps } = useProcessorDependencies(processor, processorContext, setProcessorContext);
+
+  const useCustomHooks = processor.getCustomHooks().useCustomHooks;
+  const [isLoadingCustomHooks, setIsLoadingCustomHooks] = useState(!!useCustomHooks);
+
+  const {
+    isLoadingInitialValues,
+    initialValues,
+    error: initialValuesError,
```
ibacher, line 61: Should this value be used for something?

src/components/processor-factory/form-processor-factory.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setProcessorContext((prev) => ({
+        ...prev,
+        ...{ formFields: formFieldsWithMeta },
+      }));
+    } else if (rawFormFields?.length) {
+      setProcessorContext((prev) => ({
+        ...prev,
+        ...{ formFields: rawFormFields },
+      }));
+    }
```
ibacher, line 88: Maybe something like this:

```suggestion
      setProcessorContext((prev) => merge({}, prev, formFieldAdapters ?? {}, formFieldValidators ? {}, formFieldsWithMeta?.length ? {formFields: formFieldsWithMeta} : {}, rawFormFields?.length ? { formFields: rawFormFields } : {}));
```

ibacher: `merge` from `lodash` to support deep merging, but otherwise `Object.assign()` would be the same as this with many fewer `setState()` calls.

src/provider/form-factory-provider.tsx
```diff
@@ -0,0 +1,138 @@
...
+  children,
+  formSubmissionProps,
+  setCurrentPage,
+  handleConfirmQuestionDeletion,
+}) => {
+  const rootForm = useRef<FormContextProps>();
+  const subForms = useRef<Record<string, FormContextProps>>({});
+  const layoutType = useLayoutType();
+  const { isSubmitting, setIsSubmitting, onSubmit, onError, handleClose } = formSubmissionProps;
+  const registerForm = (formId: string, context: FormContextProps, isSubForm: boolean) => {
```
ibacher, line 68: This should probably be wrapped in `useCallback()`

ibacher: I also wonder about the `isSubForm` parameter. Shouldn't the assumption be something like the first form registered is the root form and every other form is a subform, so something like:

```ts
if (!rootForm.current) {
  rootForm.current = context;
} else {
  subForm.current[formId] = context;
}
```

src/provider/form-factory-provider.tsx
```diff
@@ -0,0 +1,138 @@
...
+        formJson,
+        formProcessors: formProcessors.current,
+        layoutType,
+        workspaceLayout,
+        visit,
+        location,
+        provider,
+        registerForm,
+        getSubForms,
+        getRootForm,
```
ibacher, line 123: Is there a use case for providing these methods to children? The only use I see of them is in the "submitting" hook and it's better if we expose as minimal an API as possible to each child.

ibacher (commented):
Throwing up a few comments for you.

src/provider/form-factory-provider.tsx
samuelmale: Just updated the component to fire an abort signal when the component unmounts.

src/provider/form-factory-provider.tsx
samuelmale: The reason I used refs is that they won’t trigger a re-render on subsequent state updates. Do you want me to use the standard `useState`?

src/provider/form-factory-provider.tsx
samuelmale: Well it appears redundant for now because I'm currently hardcoding the form processors. I left myself a TODO which suggests that the processors will be promised. 

```ts
  // TODO: Manage and load processors from the registry
  const formProcessors = useRef<Record<string, FormProcessorConstructor>>({
    EncounterFormProcessor: EncounterFormProcessor,
  });

```

src/components/processor-factory/form-processor-factory.component.tsx
samuelmale: I agree that loading all the concepts is an expensive operation, and we should rely on backend optimization. I've tried loading a form using the O3 forms endpoint, but in my experience, it doesn't seem to resolve missing labels (by inferring them from the concept).

For us to fully depend on the O3 forms endpoint, we need to add support for resolving subforms.

src/provider/form-factory-provider.tsx
ibacher: A slightly different way to handle it would be to put the form processors in a separate component with a separate context and just wrap the component in `Suspense` until they're loaded, e.g.,

```tsx
<Suspense fallback={null}>
  <FormProcessorProvider>
    {children}
  </FormProcessorProvider>
</Suspense>
```

src/provider/form-factory-provider.tsx
ibacher: No, I guess since we're not sending them anywhere it's probably ok.

src/provider/form-factory-provider.tsx
samuelmale: Nice one!

src/components/processor-factory/form-processor-factory.component.tsx
ibacher: > it doesn't seem to resolve missing labels

Ah, yes, it doesn't magically create `label` elements. Basically, this was so, in the NGX engine, we could override the label with translations or, absent those, the concept name, since the concept name isn't always the appropriate string. That said, the endpoint should give you an element called "conceptReferences" which has, for every concept in the form, the concept's UUID and display name.

I guess that's a good point about subforms, but you could always just request them as separate forms for now.

src/provider/form-factory-provider.tsx
samuelmale: Nice argument! Updating 🧑‍🍳

src/components/processor-factory/form-processor-factory.component.tsx
samuelmale: This is a good solution, but it ends up losing my object references in memory due to the deep clone. This is problematic because I would have to manually reconcile the fields in the formJson tree versus the flattened fields. This can lead to increased memory usage and potential inconsistencies, as I will end up with two separate instances for each form field.

I've ended doing something like:

```ts
setProcessorContext((prev) => ({
      ...prev,
      ...(formFieldAdapters && { formFieldAdapters }),
      ...(formFieldValidators && { formFieldValidators }),
      ...(formFieldsWithMeta?.length
        ? { formFields: formFieldsWithMeta }
        : rawFormFields?.length
        ? { formFields: rawFormFields }
        : {}),
    }));

```

ibacher: Ah! Fair enough. Still... involves fewer calls to the `setProcessContext()`, which is basically the whole point.

src/components/processor-factory/form-processor-factory.component.tsx
samuelmale: Since the o3forms endpoints now supports subforms and correctly queries forms by name, I'm gonna create a ticket to track the work of fully leveraging this backend optimisation.

ibacher (approved):

PR #354, opened 2024-07-24
(fix) O3-3634: DatePicker input should clear when unspecified is checked

ODORA0:
## Requirements

## Screenshots

https://github.com/user-attachments/assets/fc416ad7-c098-434c-9ab9-2bb98948e423

## Related Issue

https://openmrs.atlassian.net/browse/O3-3634


src/components/inputs/date/date.component.tsx
```diff
@@ -97,6 +107,7 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
             <div className={styles.datePickerSpacing}>
               <Layer>
                 <OpenmrsDatePicker
+                  key={key}
```
arodidev, line 110: Could we use the field id as the key instead?

ODORA0: @arodidev When using a dynamic key like `field.value`, the component may not behave as expected in some cases. So using the original approach which also ensure the component updates correctly when unspecified is check works well with tests as well

samuelmale:
@ODORA0 did you see the CI failure?

src/components/inputs/date/date.component.tsx
```diff
@@ -82,7 +84,15 @@ const DateField: React.FC<FormFieldProps> = ({ question, onChange, handler, prev
     }
   }, [field.value, time]);
 
-  return encounterContext.sessionMode == 'view' || encounterContext.sessionMode == 'embedded-view' ? (
+  useEffect(() => {
+    if (unspecifiedField.value) {
+      setFieldValue(question.id, null);
+      setTime(null);
+      setKey(prevKey => prevKey + 1);
```
NethmiRodrigo, line 91: Why do we use an int as key as increment it?

src/components/inputs/date/date.component.tsx
NethmiRodrigo: The question id is not dynamic, and would be much more suited rather than the int here

src/components/inputs/date/date.component.tsx
NethmiRodrigo: CC @samuelmale

src/components/inputs/date/date.component.tsx
ODORA0: The major issue is with how the OpenmrsDatePicker propagates the states from the main component to the DatePickerInput component in esm core, when the value is cleared it never trickles down cc @samuelmale
We can report this bug but doing this as well will also re-render the component thus clearing the input when unspecified is check

samuelmale: @ibacher any thoughts on the above?

brandones:
@ODORA0 @samuelmale @NethmiRodrigo Is it clear what needs to happen before we can put this in? I've fixed the merge conflicts. Looks like CI is passing. Can someone please file a ticket for the bug where the datepicker doesn't update when its value is cleared?

NethmiRodrigo:
@brandones I think we might not need this or need to change this following the fix that @ibacher made in this [PR](https://github.com/openmrs/openmrs-esm-core/pull/1099). I made a separate PR to update the datepicker but trying to get the tests fixed before that

brandones:
@NethmiRodrigo does that PR supersede this one? If so please close this PR and link O3-3634 to your PR.

NethmiRodrigo:
@brandones doesn't exactly supersede, I just updated the onChange function, but I could bundle in the fix for this in that PR

brandones:
@NethmiRodrigo if the solution in this PR will still be the right solution, then just merge yours and then we'll merge this PR. If this PR no longer has anything useful about it, just add the solution to your PR and close this one.

brandones:
(or open a new PR, however you like)

src/components/inputs/date/date.component.tsx
ibacher: This should be fixed.

samuelmale:
Closing this because it was addressed [here](https://github.com/openmrs/openmrs-esm-core/pull/1099).

PR #367, opened 2024-08-20
(fix) parse NumberInput value to Number

kajambiya:
## Requirements

## Summary
The value from the NumberInput field was being captured as a string. As a result, the form-validator would return false when checking if the value is an integer. 

## Screenshots
Before fix:

https://github.com/user-attachments/assets/8ba2341a-ac9b-48ea-8932-6dec56cec057

After fix:

https://github.com/user-attachments/assets/9244109b-2e4c-4a4f-b1e1-11baeb172d1f

### DIFF
--- a/src/components/inputs/number/number.component.tsx
+++ b/src/components/inputs/number/number.component.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { Layer, NumberInput } from '@carbon/react';
 import classNames from 'classnames';
 import { isTrue } from '../../../utils/boolean-utils';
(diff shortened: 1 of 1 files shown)


src/components/inputs/number/number.component.tsx
```diff
@@ -22,6 +22,11 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
     }
   };
 
+  const handleChange = (event) => {
```
samuelmale, line 25: Not a big deal but can we wrap this in a `useCallback(..)`?

src/components/inputs/number/number.component.tsx
```diff
@@ -22,6 +22,11 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
...
+    const parsedValue = parseFloat(event.target.value);
```
samuelmale, line 26:
```suggestion
    const parsedValue = Number(event.target.value);
```

src/components/inputs/number/number.component.tsx
kajambiya: Done

samuelmale (approved):

PR #369, opened 2024-08-28
(fix) Resolve duplicate form registration

samuelmale:
## Requirements

## Summary

Added a check to the `registerForm` function to verify if the incoming form is not already set as the root form.

## Screenshots

N/A
## Related Issue

https://openmrs.atlassian.net/browse/O3-3882

### DIFF
--- a/src/provider/form-factory-provider.tsx
+++ b/src/provider/form-factory-provider.tsx
@@ -83,7 +83,7 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
   const registerForm = useCallback((formId: string, context: FormContextProps) => {
     if (!rootForm.current) {
       rootForm.current = context;
-    } else {
+    } else if (rootForm.current.formJson.name !== formId) {
       subForms.current[formId] = context;
     }
   }, []);


src/provider/form-factory-provider.tsx
```diff
@@ -83,7 +83,7 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
   const registerForm = useCallback((formId: string, context: FormContextProps) => {
     if (!rootForm.current) {
       rootForm.current = context;
-    } else {
+    } else if (rootForm.current.formJson.name !== formId) {
```
pirupius, line 86: Doesn't `formId` have the uuid of the form? If that's the case then this might be wrong because the comparison seems to be with the form `name`

samuelmale: It's currently based on [the form name](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/components/renderer/form/form-renderer.component.tsx#L67) (this was for troubleshooting purposes)

pirupius (approved):

PR #373, opened 2024-09-05
(fix) Reactively cascade form field update event

samuelmale:
## Requirements

## Summary

**Background:** In the form-engine, state management is typically handled using a reducer pattern, often in conjunction with React's useReducer hook. This is propped with helpers to conveniently update the state eg. [updateFormField](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/hooks/useFormStateHelpers.ts#L9-L11). 

**Issue:** The form-renderer currently [receives the field object directly from the section](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/components/renderer/section/section-renderer.component.tsx#L15) instead of the reactive flattened form fields state. This obscures the cascade of the reactive events.

**Solution**: Treat the flattened `formFields` state as the source of truth.

## Screenshots

N/A

## Related Issue

https://openmrs.atlassian.net/browse/O3-3908

### DIFF
--- a/src/components/group/obs-group.component.tsx
+++ b/src/components/group/obs-group.component.tsx
@@ -16,7 +16,7 @@ export const ObsGroup: React.FC<FormFieldInputProps> = ({ field }) => {
         return (
           <div className={classNames(styles.flexColumn)} key={keyId}>
             <div className={styles.groupContainer}>
-              <FormFieldRenderer field={child} valueAdapter={formFieldAdapters[child.type]} />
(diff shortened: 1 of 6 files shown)


samuelmale:
So this enables us to effectively subscribe to form field update events:

```tsx
const items = useMemo(() => {
    // logic
    return [];
}, [field.questionOptions.answers]);

return <>...
```

pirupius (approved):
LGTM

PR #390, opened 2024-09-11
O3 3685: Enable Deleting or Clearing Files in File Picker Component in Edit Mode

TrevorAntony:
## Requirements
Enable Deleting or Clearing Files in File Picker Component in Edit Mode

## Summary
When a file is selected using the file picker component and the form is later opened in edit mode, there is now an option to delete or clear the selected file. 

## Screenshots
Screenshot 2024-09-11 at 13 16 23
Screenshot 2024-09-11 at 13 16 31

## Related Issue
https://openmrs.atlassian.net/browse/O3-3685

### DIFF
--- a/src/components/inputs/file/file.component.tsx
+++ b/src/components/inputs/file/file.component.tsx
@@ -3,7 +3,7 @@ import { FileUploader, Button } from '@carbon/react';
 import { useTranslation } from 'react-i18next';
 import { isTrue } from '../../../utils/boolean-utils';
 import Camera from './camera/camera.component';
-import { Close, DocumentPdf } from '@carbon/react/icons';
+import { Close, DocumentPdf, TrashCan } from '@carbon/react/icons';
 import styles from './file.scss';
 import { type FormFieldInputProps } from '../../../types';
 import { useFormProviderContext } from '../../../provider/form-provider';
@@ -49,6 +49,12 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
     },
     [setFieldValue],
   );
+
+  const handleDeleteFile = useCallback(() => {
+    setImagePreview(null);
+    setFieldValue(null);
+    setDataSource(null);
+  }, [setFieldValue]);
 
   if (isViewMode(sessionMode) && !value) {
     return (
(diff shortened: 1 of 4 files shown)


src/components/inputs/file/file.component.tsx
```diff
@@ -49,6 +49,13 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
     },
     [setFieldValue],
   );
+  const [clearedFiles, setClearedFiles] = useState([]);
```
samuelmale, line 52: Can you move this state declaration at the top? 

Question: Are you using the "clearedFiles" state at any point?

src/components/inputs/file/file.component.tsx
```diff
@@ -49,6 +49,13 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
...
+  const handleClearFile = (fileId) => {
+    setClearedFiles((prevClearedFiles) => [...prevClearedFiles, fileId]);
+    setFieldValue(null);
+    setImagePreview(null);
+    setDataSource(null);
+  };
```
samuelmale, line 58: Can you wrap this handler wrapped in a useCallback?

samuelmale (changes requested):
Thanks for working on this @TrevorAntony; Just a few general concerns:

1. Is the "delete" operation persistent on form submission? Do we void the complex obs?
2. Can you add unit test(s) to assert that the cleared obs gets voided?

src/components/inputs/file/file.component.tsx
arodidev: @TrevorAntony I don't think we need to maintain the state of the cleared values, resetting the other three should suffice.

```ts
const handleClearFile = useCallback(() => {
  setFieldValue(null);
  setImagePreview(null);
  setDataSource(null);
}, []);
```

src/components/inputs/file/file.component.tsx
TrevorAntony: noted

src/components/inputs/file/file.component.tsx
TrevorAntony: yes done this

TrevorAntony:
Having an issue where even after voiding/clearing the obs and successful form submission, when you relaunch the form in  edit mode, the cleared image/file is still displayed

PR #393, opened 2024-09-13
(Enhancement) Correctly handle dateValue with null or invalid date

ODORA0:
## Requirements

## Summary
The core issue was that the dateValue being passed to the OpenmrsDatePicker and TimePicker components was often null or an invalid date (which shows up as Invalid Date in JavaScript). When these components received such values, they couldn’t handle them properly, which caused the RangeError: Invalid time value 

**Solution:**
Validation with isValidDate:

I introduced the isValidDate helper function to ensure that only valid Date objects are processed. This function checks if the `dateValue` is a valid instance of Date and if it's not NaN (which happens when a Date is constructed with invalid input).
Wherever we were passing dateValue to the components, we first check if it’s valid. If it's not, we provide fallback behavior, like setting the value to null or creating a new valid Date where necessary.
Fallback Values for Invalid Dates:

In the components, we now handle null or invalid dates gracefully by either defaulting to a valid date (like new Date()) or setting the value to null in case of an invalid date. This prevents the components from attempting to format invalid date values.
Conditional Rendering and Disabling:

The TimePicker is disabled unless there's a valid date (dateValue). This ensures that we don’t allow interaction with the time input unless the user has selected a valid date.
We also ensure that invalid dates are not passed into the TimePicker or DatePicker components to avoid any further rendering errors.

## Screenshots
image

image

### DIFF
--- a/src/components/inputs/date/date.component.tsx
+++ b/src/components/inputs/date/date.component.tsx
@@ -12,6 +12,9 @@ import { OpenmrsDatePicker, formatDate, formatTime } from '@openmrs/esm-framewor
 import { useFormProviderContext } from '../../../provider/form-provider';
 import FieldLabel from '../../field-label/field-label.component';
 
+// Helper function to validate date
+const isValidDate = (date) => date instanceof Date && !isNaN(date.getTime());
+
 const DateField: React.FC<FormFieldInputProps> = ({ field, value: dateValue, errors, warnings, setFieldValue }) => {
   const { t } = useTranslation();
   const [time, setTime] = useState('');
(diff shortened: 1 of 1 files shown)


samuelmale:
@ODORA0 are you still working this?

ODORA0:
@samuelmale If this was resolved I can close it otherwise I can clean this and open it for review.

samuelmale:
> If this was resolved I can close it

@ODORA0 can you show me how I can reproduce the issue?

ODORA0:
@samuelmale Try [this form](https://pastebin.com/Ab7cYP7W), you will notice that on the next visit date fails to render. But when you remove the calculate expression it renders. However the idea behind the expression is that it should provide a date in the next visit date field 30 days after the date provided in the Registration Date field above it. This PR intended to fix that

ODORA0:
> I can reproduce the issue?

samuelmale:
I get the issue; in your case, the helper function is returning an invalid date. How about fixing the helper function?

PR #394, opened 2024-09-16
(Enhancement): O3-3685: Enable Deleting or Clearing Files in File Picker Component in Edit Mode

TrevorAntony:
## Requirements
When a file is selected using the file picker component and the form is later opened in edit mode, there is currently no option to delete or clear the selected file. 

## Summary
The user can now successfully edit an encounter and add or remove an image/file by clearing it.

## Screenshots
Screenshot 2024-09-16 at 11 15 59
Screenshot 2024-09-16 at 11 15 49

## Related Issue
https://openmrs.atlassian.net/browse/O3-3685

## Other
the delete persists on form submission and reloading

### DIFF
--- a/src/components/inputs/file/file.component.tsx
+++ b/src/components/inputs/file/file.component.tsx
@@ -3,13 +3,14 @@ import { FileUploader, Button } from '@carbon/react';
 import { useTranslation } from 'react-i18next';
 import { isTrue } from '../../../utils/boolean-utils';
 import Camera from './camera/camera.component';
-import { Close, DocumentPdf } from '@carbon/react/icons';
+import { Close, DocumentPdf, TrashCan } from '@carbon/react/icons';
 import styles from './file.scss';
 import { type FormFieldInputProps } from '../../../types';
 import { useFormProviderContext } from '../../../provider/form-provider';
 import { isViewMode } from '../../../utils/common-utils';
 import FieldValueView from '../../value/view/field-value-view.component';
 import FieldLabel from '../../field-label/field-label.component';
+import { type Attachment, deleteAttachmentPermanently, showSnackbar } from '@openmrs/esm-framework';
 
 type DataSourceType = 'filePicker' | 'camera' | null;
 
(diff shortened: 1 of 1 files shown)


src/components/inputs/file/file.component.tsx
```diff
@@ -87,6 +111,13 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
             {t('cameraCapture', 'Camera capture')}
           </Button>
         </div>
+        {value && (
```
samuelmale, line 114: This shouldn't be available in "view mode";

```tsx
{value &&  !isViewMode(sessionMode) && (

```

(Import `isViewMode(mode)` from `../utils/common-utils`)

src/components/inputs/file/file.component.tsx
```diff
@@ -87,6 +111,13 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
...
+          <div className={`${styles.selectorButton} ${styles.clearFileButton}`}>
```
samuelmale, line 115: Does the `clearFileButton` class exist in `./file.scss`?

src/components/inputs/file/file.component.tsx
```diff
@@ -20,6 +21,29 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
   const [dataSource, setDataSource] = useState<DataSourceType>(null);
   const { sessionMode } = useFormProviderContext();
 
+  const deleteAttachment = useCallback(() => {
```
samuelmale, line 24: I would name this something like: `handleDeleteAttachment`

src/components/inputs/file/file.component.tsx
```diff
@@ -20,6 +21,29 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
...
+    if (value && value.id) {
+      deleteAttachmentPermanently(value.id, new AbortController())
```
samuelmale, line 26: The abort controller isn't being correctly used here (I know we keep making this same mistake across the codebase). You want to emit an abort signal when the component unmounts.

```tsx

// declare controller
const abortController = new AbortController();

useEffect(() => {
  // emit an abort signal when the component unmounts 
   return () => abortController.abort();
}, []);

const handleDeleteAttachment =  useCallback(() => {  
   // logic that consumes the controller

}, [value, setFieldValue, t]);

```

src/components/inputs/file/file.component.tsx
```diff
@@ -20,6 +21,29 @@ const File: React.FC<FormFieldInputProps> = ({ field, value, setFieldValue }) =>
...
+          showSnackbar({
+            title: t('error', 'Error'),
+            subtitle: t('failedDeleting', "File couldn't be deleted"),
+            kind: 'error',
+          });
+        });
+    } else {
+      setFieldValue(null);
+    }
+  }, [value, setFieldValue, t]);
```
samuelmale, line 46: Can you add a line break?

samuelmale (changes requested):

src/components/inputs/file/file.component.tsx
samuelmale: @TrevorAntony a better way of approaching the above is by using [useAbortController](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-react-utils/src/useAbortController.ts#L25) defined in core:


```tsx

  const abortController = useAbortController();

  const handleDeleteAttachment =  useCallback(() => {  
     // logic that consumes the controller
  
  }, [value, setFieldValue, t]);

```

arodidev:
Thanks for this @TrevorAntony. LGTM.

samuelmale:
@TrevorAntony The "Clear File" button seems to be stretched and overlapped by the side rail. I think we should have an actions menu associated with the file itself, or we could simply place the delete button next to the attachment. This is important considering that we plan to support capturing multiple attachments. Feel free to [ask for some UX advisory](https://openmrs.slack.com/archives/CKS32D55G) if necessary.


Screenshot 2024-09-25 at 18 56 16

samuelmale:
@TrevorAntony are you still working on this?

brandones:
@TrevorAntony Are you planning to pick this back up, and if so, when?

PR #407, opened 2024-10-04
(fix) Evaluate calculate expressions after general form field initialization

pirupius:
## Requirements

## Summary
This PR ensures that the form engine evaluates calculate expressions after the general form field initialization. This ensures that any referenced fields (in the calculate expression) are initialised prior.  

Before:
Screenshot 2024-10-04 at 15 26 36

Applying any logic would default it to 1970 because of the null value

Screenshot 2024-10-04 at 15 26 18

After: 
Screenshot 2024-10-04 at 15 26 56

Screenshot 2024-10-04 at 15 24 00

### DIFF
--- a/src/components/inputs/date/date.component.tsx
+++ b/src/components/inputs/date/date.component.tsx
@@ -61,7 +61,7 @@ const DateField: React.FC<FormFieldInputProps> = ({ field, value: dateValue, err
         setTime([hours, minutes].join(':'));
       }
     }
-  }, [dateValue]);
+  }, []);
 
   const timePickerLabel = useMemo(
     () =>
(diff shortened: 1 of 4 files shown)


src/types/schema.ts
```diff
@@ -215,8 +215,6 @@ export type RenderType =
   | 'toggle'
   | 'ui-select-extended'
   | 'workspace-launcher'
-  | 'fixed-value'
-  | 'file'
```
pirupius, line 219: These were duplicated

src/components/inputs/date/date.component.tsx
```diff
@@ -54,6 +54,7 @@ const DateField: React.FC<FormFieldInputProps> = ({ field, value: dateValue, err
 
   useEffect(() => {
     if (dateValue) {
+      setFieldValue(dateValue);
```
pirupius, line 57: On initialization, the date field is set to the current date but this value wasn't being set so validation and reusing the value in other parts of the form wasn't possible because it was null

samuelmale (approved):

PR #417, opened 2024-10-24
(feat) Restore side navigation

samuelmale:
## Requirements

## Summary

Brings back the form's side navigation with some improvements! The nav shows up when you have multiple pages and enough screen space (extra-wide or ultra-wide workspace).

### Under the Hood
Three new hooks make this work:
1. `useFormWorkspaceSize`: Figures out how much space we have to work with
2. `usePageObserver`: Keeps track of pages, errors, and what's visible
3. `useCurrentActivePage`: Handles page navigation 

### Known Limitations
Few things we might want to add later:
- Can't jump directly to fields with errors yet
- No toggle to hide/show the navigation
- No "mark all as unspecified" feature

## Screenshots

extra-wide workspace:

ultra-wide workspace:

Screenshot 2024-10-29 at 14 57 21

wider workspace:

Screenshot 2024-10-29 at 15 00 36


src/components/sidebar/sidebar.component.tsx
```diff
@@ -1,134 +1,91 @@
...
+                onClick={(e) => {
+                  e.preventDefault();
+                  requestPage(page.id);
+                }}>
+                <span>{page.label}</span>
+              </button>
+            </div>
+          );
+        })}
+      {sessionMode !== 'view' && <hr className={styles.divider} />}
```
pirupius, line 62: There might be a use case for `embedded-view` mode within the active visits that might necessitate this being updated.

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+import { scrollIntoView } from '../../utils/form-helper';
+
+interface UseCurrentActivePageProps {
+  pages: FormPage[];
+  defaultPage: string;
+  activePages: string[];
+  evaluatedPagesVisibility: boolean;
+}
+
+interface UseCurrentActivePageResult {
```
pirupius, line 12: Adding `Use***` for interface naming makes the naming convention confusing. I think we should just change these 2 to `CurrentActivePageProps` and `CurrentActivePageResult` respectively especially since there not being exported anyways.

src/form-engine.component.tsx
```diff
@@ -56,35 +55,40 @@ const FormEngine = ({
...
+
+    return ['narrow', 'wider'].includes(workspaceSize) || !hasMultiplePages;
+  }, [mode, workspaceSize, isLoadingDependencies, hasMultiplePages]);
+
+  const showSidebar = useMemo(() => {
+    if (mode === 'embedded-view' || isLoadingDependencies || hasMultiplePages === null) {
+      return false;
+    }
+
+    return ['extra-wide', 'ultra-wide'].includes(workspaceSize) && hasMultiplePages;
```
pirupius, line 90: These modes are starting to get confusing. This implies 4 modes yet the design guidelines as well as workspace only has 3 modes. Does `extra-wide` have a different behavour from `ultra-wide`

src/hooks/useFormWorkspaceSize.ts
```diff
@@ -0,0 +1,51 @@
...
+import { pxToRem } from '../utils/common-utils';
+
+/**
+ * The width of the supported workspace variants in rem
+ */
+const narrowWorkspaceWidth = 26.25;
+const widerWorkspaceWidth = 32.25;
+const extraWideWorkspaceWidth = 48.25;
+
+type WorkspaceSize = 'narrow' | 'wider' | 'extra-wide' | 'ultra-wide';
```
pirupius, line 11: This indicates that `extra-wide` and `ultra-wide` might be doing the same thing so maybe we should drop `extra-wide` naming

src/hooks/useFormWorkspaceSize.ts
```diff
@@ -0,0 +1,51 @@
...
+  const size = useMemo(() => {
+    if (containerWidth <= narrowWorkspaceWidth) {
+      return 'narrow';
+    } else if (containerWidth <= widerWorkspaceWidth) {
+      return 'wider';
+    } else if (containerWidth <= extraWideWorkspaceWidth) {
+      return 'extra-wide';
+    } else {
+      return 'ultra-wide';
+    }
```
pirupius, line 28: 🤯 which dimensions will `ultra-wide` take up? or will it inherit from the screen size? This still feels like unnecessary.

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+        const firstVisiblePage = pages.find((page) => !page.isHidden);
+        if (firstVisiblePage) {
+          setCurrentActivePage(firstVisiblePage.id);
+        }
+      }
+    };
+
+    initializePage();
+    setIsInitialized(true);
+  }, [pages, defaultPage, evaluatedPagesVisibility, isInitialized]);
```
ibacher, line 72: Why bother with the `useEffect()` here at all?:

```ts
if (!isInitialized && evaluatedPagesVisibility) {
  const defaultPageObject = pages.find(({ label }) => label === defaultPage);
  // blah, blah, blah
}
```

This is actually also a weird case where using a `ref` instead of state to track whether or not the hook is initialized is actually a performance improvement because the side effect of calling `setIsInitialized()` is actually undesirable (we only want the re-render if `setCurrentActivePage()` is called.

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  useEffect(() => {
+    // Lock out Waypoint updates for 200ms to allow for:
+    // 1. Initial render completion
+    // 2. Scroll position establishment
+    // 3. Waypoint to complete its initial visibility detection
+    if (isInitialized) {
```
ibacher, line 79: AFAICT, `initialLockTimeoutRef` is only used locally to this hook, so why not have it as a local?

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+      initialLockTimeoutRef.current = setTimeout(() => {
+        isInitialPhaseRef.current = false;
+      }, 200);
+    }
+
+    // Cleanup
+    return () => {
```
ibacher, line 86:
```suggestion
    return initialLockTimeoutRef.current ? () => {
      // snip
    }) : null;
```

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Handle active pages updates from viewport visibility
+  useEffect(() => {
+    if (isInitialPhaseRef.current) return;
+
+    const updateActivePage = () => {
+      // If there's a requested page and it's visible, keep it active
+      if (requestedPage && activePages.includes(requestedPage)) {
+        setCurrentActivePage(requestedPage);
+        setTimeout(() => {
```
ibacher, line 99: If we're creating a timeout, we should clear the timeout when the effect returns.

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+          setRequestedPage(null);
+        }, 100);
```
ibacher, line 101: Why 100ms? Could this just be 0?

src/components/sidebar/useCurrentActivePage.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // Cleanup
+    return () => {
+      clearTimeout(initialLockTimeoutRef.current);
+    };
+  }, [isInitialized]);
+
+  // Handle active pages updates from viewport visibility
+  useEffect(() => {
```
ibacher, line 92: Is there a way we can move this logic into the `requestPage()` callback itself?

src/components/sidebar/page-observer.ts
```diff
@@ -0,0 +1,58 @@
+import { BehaviorSubject } from 'rxjs';
+import { type FormPage } from '../../types';
+
+class PageObserver {
```
ibacher, line 4: A couple of points here:

1. I don't love that this is a library-wide singleton rather than an object tied to a specific render tree, which is what it feels like it should be. This basically adds the artificial restriction that only one form instance (per instance of form-engine-lib) is active at a time.
2. Since we have a `usePageObserver()` hook, wouldn't it be better to confine all the reactivity to that? I'm not clear on the advantages of having a separate class here with a bunch of BehaviorSubjects.

src/components/sidebar/sidebar.component.tsx
```diff
@@ -1,134 +1,91 @@
...
+              <button
+                onClick={(e) => {
+                  e.preventDefault();
+                  requestPage(page.id);
+                }}>
+                <span>{page.label}</span>
+              </button>
+            </div>
+          );
+        })}
```
ibacher, line 61: This feels like the type of thing that might be better handled in a `useMemo()` hook. At the _least_, the top-level divs here need a `key` property.

ibacher: Better yet, we would refactor the div / button combo into its own component (it can be an internal component local to this file).

src/components/sidebar/sidebar.scss
```diff
@@ -1,102 +1,87 @@
...
-  color: colors.$gray-100;
-  cursor: pointer;
-
-  :hover {
-    outline: none;
-  }
-}
-
-.section {
+.tab {
```
ibacher, line 4: I think `.tab` is a pretty bad name here. Surely "page" or "pageLink" is better?

ibacher (commented):
Nice work! A few nit-picking comments.

src/form-engine.component.tsx
ibacher: `ultra-wide` is basically "full screen". See the screenshots in the PR.

pirupius: Thanks @ibacher i saw that. Then i guess the issue would be with `extra-wide` unless that naming is being referenced else where within the patient chart.

src/components/sidebar/page-observer.ts
samuelmale: Nice catch!

src/components/sidebar/useCurrentActivePage.ts
ibacher: Ah, yes. When I wrote this, I was hoping to get rid of most of the `useEffect()`s, but we definitely need it for the timeout

src/components/sidebar/sidebar.component.tsx
samuelmale: >This feels like the type of thing that might be better handled in a useMemo() hook.

A more ideal approach would be memoizing this once in the form renderer and we expect the page observer to report only visible pages, but there is an issue:

Achieving this is likely to beg for the need of a state update of the entire form object every time a field changes, triggering the form to re-render every time the user interacts with a field. Currently, when a field's state changes, only the field and the flattened fields array state are updated (this doesn't trigger a re-render of the entire form).

But why would this be necessary? This boils down to how we determine the visibility of a page. A page is considered to have visible content if:

- The page itself is not hidden.
- At least one section within the page is visible.
- At least one question within each section is visible.

I think for now, we can deploy keys to the heat-map level elements, leveraging React's reconciliation algorithm and address the bigger issue in a different PR.

src/components/sidebar/useCurrentActivePage.ts
samuelmale: It's possible but that may introduce some coupling or confusion because these events are propagated by disparate actions. A page request is made when the user clicks on a link in the sidenav while the other events are coming from Waypoint.

src/components/sidebar/useCurrentActivePage.ts
samuelmale: This pushes the cleanup or clearing of the requested page to a later point. This ensures that we ignore all intermediate Waypoint events as we scroll to the requested page. So the 100ms delay is necessary.

src/components/sidebar/sidebar.component.tsx
ibacher: Well, yeah, the whole need to re-render the form was kind of why I was suggesting local memoization, but I think I see your point: any field update can (potentially) trigger new fields to show and effect page visibility, and there isn't an obvious way to tie memoization to changing page visibility, so let's leave that aside.

src/hooks/useFormWorkspaceSize.ts
```diff
@@ -0,0 +1,51 @@
...
+  }, [containerWidth]);
+
+  useLayoutEffect(() => {
+    const handleResize = () => {
+      const containerWidth = rootRef.current?.parentElement?.offsetWidth;
+      containerWidth && setContainerWidth(pxToRem(containerWidth));
```
ibacher, line 34: So, for this to be done correct, we need to always pass the actual root element font size:

```ts
const rootFontSize = parseInt(getComputedStyle(document.documentElement).fontSize);
```

src/hooks/useFormWorkspaceSize.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+    handleResize();
+    const resizeObserver = new ResizeObserver((entries) => {
+      handleResize();
+    });
+
+    if (rootRef.current) {
```
ibacher, line 41: Shouldn't the check here actually be:
```suggestion
    if (rootRef.current?.parentElement) {
```

src/hooks/useFormWorkspaceSize.ts
samuelmale: Nice catch!

src/components/sidebar/sidebar.component.tsx
samuelmale: I'm not sure of the use case but currently, the sidenav isn't supported in "embedded-view"

samuelmale:
@ibacher kindly review my latest changes so that we can have this go in.

ibacher (approved):

PR #419, opened 2024-10-25
(refactor) FormField attribute 'Type' should be Optional

Twiineenock:
## Requirements

## Summary

Markdown questions have no question `type` hence should be optional in the form `schema`.
This related to the work coming in from here [We should make this optional since markdown content doesn't have a type](https://github.com/openmrs/openmrs-esm-form-builder/pull/355#discussion_r1816594802)

### DIFF
--- a/src/types/schema.ts
+++ b/src/types/schema.ts
@@ -58,7 +58,7 @@ export interface FormSection {
 
 export interface FormField {
   label?: string;
-  type: string;
+  type?: string;
   questionOptions: FormQuestionOptions;
   datePickerFormat?: 'both' | 'calendar' | 'timer';
   id: string;


NethmiRodrigo (approved):
LGTM! @samuelmale we had to make type optional as well. Does that seem okay?

samuelmale:
This makes sense and I'm not opposed to the idea. 

> Markdown questions have no question type hence should be optional in the form schema

The FE defines a pseudo type called `control` ([see](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/adapters/control-adapter.ts)) and I believe it should be used in such scenarios. If this makes sense, we can have this type as a reasonable default in the form-builder. cc: @denniskigen

```json
           {
             "type": "control"
              "value": [
                "**This form is used to**:showcase feature of the react form engine",
                "### This is just an **explainer text** example  in *markdown*"
              ],
              "questionOptions": {
                "rendering": "markdown"
              },
              "id": "heading2"
            }
```

denniskigen:
Seems reasonable to me, @samuelmale!

NethmiRodrigo:
@samuelmale that makes sense! Shall we opt to do that instead then, default `type` to the value `control`, without making it optional? FYI @Twiineenock

NethmiRodrigo:
Closing this PR in favor of making the `type` to `control` for markdown content

PR #435, opened 2024-12-02
(chore) Bump @carbon/react to v1.71.0

elishabantana:
## Requirements

## Summary
This PR updated the version of @carbon/react to the latest, v1.71.0 in esm-form-engine-lib.

## Related Issue
https://openmrs.atlassian.net/browse/O3-4229

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@carbon/react": ">1.47.0 <1.50.0",
+    "@carbon/react": "1.71.0",
     "classnames": "^2.5.1",
     "lodash-es": "^4.17.21",
     "react-error-boundary": "^4.0.13",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1280,6 +1280,15 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@carbon/react": ">1.47.0 <1.50.0",
```
pirupius, line 32: Hi @elishabantana there's an ongoing discussion to make this a peer dependency all together. The reason it was previously like this was because of Carbon's searchable multi-select behavior between versions. 
We also wouldn't want to give a static version because this is package is only a library and not a microfrontend

elishabantana:
So how do I bump its carbon version to the latest

package.json
NethmiRodrigo: @pirupius sorry I'm a bit lost. Does that mean we shouldn't update carbon or we should update it to be a more up-to-date range?

pirupius: @NethmiRodrigo The reason this was added was to resolve a breaking change in some components. The solution would involve
1. Testing and fixing any breaking changes in some components with newer versions Carbon 
2. Move carbon back to being a peer dependency.

PR #440, opened 2024-12-04
(feat) O3-3911 Display all items on ui-select-extended

ODORA0:
## Requirements

## Summary
- Updated shouldFilterItem to always return true, ensuring all items are displayed when the caret is clicked.
- Adjusted tests to validate that all items remain visible even after typing input.

## Screenshots

https://github.com/user-attachments/assets/85e7989e-39fc-44df-9876-0f7bfd686a2d

## Related Issue
https://openmrs.atlassian.net/browse/O3-3911

### DIFF
--- a/src/components/inputs/ui-select-extended/ui-select-extended.component.tsx
+++ b/src/components/inputs/ui-select-extended/ui-select-extended.component.tsx
@@ -148,13 +148,7 @@ const UiSelectExtended: React.FC<FormFieldInputProps> = ({ field, errors, warnin
             itemToString={(item) => item?.display}
             selectedItem={selectedItem}
             placeholder={isSearchable ? t('search', 'Search') + '...' : null}
-            shouldFilterItem={({ item, inputValue }) => {
-              if (!inputValue) {
-                // Carbon's initial call at component mount
-                return true;
-              }
-              return item.display?.toLowerCase().includes(inputValue.toLowerCase());
-            }}
+            shouldFilterItem={() => true}
             onChange={({ selectedItem }) => {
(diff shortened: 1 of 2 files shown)


samuelmale:
Can you update your branch to pick up [the fix](https://github.com/openmrs/openmrs-esm-form-engine-lib/pull/441) for the failing CI build?

samuelmale:
@ODORA0 is it possible to fix the [issue](https://openmrs.atlassian.net/browse/O3-3911) without getting rid the filter?

ODORA0:
@samuelmale The filtering restricts options because the `shouldFilterItem` function dynamically excludes items that don’t match the current `inputValue`. If the logic doesn’t explicitly account for an empty inputValue (e.g. when the caret is clicked), only the currently selected item or no items may be displayed, as the filter prevents non-matching items from appearing.

denniskigen:
Thanks, @ODORA0. How different is this than getting rid of the prop altogether? AFAICT, the filtering behaviour of the ComboBox input where only the matching item is listed when the dropdown is clicked after selecting an item is by design. See the underlying useComboBox hook implementation for the Downshift library [here](https://www.downshift-js.com/use-combobox). I think it's preferable to just remove the prop altogether until the underlying behavior changes upstream from Carbon.

ODORA0:
@denniskigen I agree, change made

samuelmale:
Ping @denniskigen

denniskigen:
Good to go for me same

pirupius (approved):
Thanks @ODORA0

PR #454, opened 2024-12-21
(feat) O3-4201: Enhance Number Question Labels Display Unit and Range (Min/Max) from Concept

D-matz:
Add (unit) (min-max) to label, validate input against concept min/max

https://openmrs.atlassian.net/browse/O3-4201

Related: https://openmrs.atlassian.net/browse/O3-4122 

4122 also suggests a warning for values within the absolute range but outside normal or critical range. Looking for suggestions on how to implement that, and how to add a test case for a concept with units and min/max

## Requirements

### DIFF
--- a/src/hooks/useConcepts.ts
+++ b/src/hooks/useConcepts.ts
@@ -5,7 +5,7 @@ import { type FetchResponse, type OpenmrsResource, openmrsFetch, restBaseUrl } f
 type ConceptFetchResponse = FetchResponse<{ results: Array<OpenmrsResource> }>;
 
 const conceptRepresentation =
-  'custom:(uuid,display,conceptClass:(uuid,display),answers:(uuid,display),conceptMappings:(conceptReferenceTerm:(conceptSource:(name),code)))';
+  'custom:(units,lowAbsolute,hiAbsolute,uuid,display,conceptClass:(uuid,display),answers:(uuid,display),conceptMappings:(conceptReferenceTerm:(conceptSource:(name),code)))';
 
 export function useConcepts(references: Set<string>): {
   concepts: Array<OpenmrsResource> | undefined;
--- a/src/hooks/useFormFieldsMeta.ts
+++ b/src/hooks/useFormFieldsMeta.ts
@@ -11,6 +11,20 @@ export function useFormFieldsMeta(rawFormFields: FormField[], concepts: OpenmrsR
         const matchingConcept = findConceptByReference(field.questionOptions.concept, concepts);
         field.questionOptions.concept = matchingConcept ? matchingConcept.uuid : field.questionOptions.concept;
         field.label = field.label ? field.label : matchingConcept?.display;
+        if(matchingConcept)
+        {
+          if(matchingConcept.units)
+          {
+            field.label = field.label + " (" + matchingConcept.units + ")";
+          }
+          if(matchingConcept.lowAbsolute && matchingConcept.hiAbsolute)
+          {
+            field.label = field.label + " (" + matchingConcept.lowAbsolute + "-" + matchingConcept.hiAbsolute + ")";
+            field.questionOptions.min = matchingConcept.lowAbsolute;
+            field.questionOptions.max = matchingConcept.hiAbsolute;
+          }
+        }
+
         if (
           codedTypes.includes(field.questionOptions.rendering) &&
           !field.questionOptions.answers?.length &&


D-matz:
@samuelmale @ibacher please let me know if you had other approaches in mind for this or issues
lower
within

greater

src/hooks/useConcepts.ts
```diff
@@ -5,7 +5,7 @@ import { type FetchResponse, type OpenmrsResource, openmrsFetch, restBaseUrl } f
 type ConceptFetchResponse = FetchResponse<{ results: Array<OpenmrsResource> }>;
 
 const conceptRepresentation =
-  'custom:(uuid,display,conceptClass:(uuid,display),answers:(uuid,display),conceptMappings:(conceptReferenceTerm:(conceptSource:(name),code)))';
+  'custom:(units,lowAbsolute,hiAbsolute,uuid,display,conceptClass:(uuid,display),answers:(uuid,display),conceptMappings:(conceptReferenceTerm:(conceptSource:(name),code)))';
```
samuelmale, line 8: @ibacher when is it necessary to use `hiCritical` vs `lowCritical`?

samuelmale (commented):
Can you add some test coverage asserting that the units and labels are incorporated in the label?

src/hooks/useConcepts.ts
ibacher: All of the values that aren't `lowAbsolute` and `hiAbsolute` are basically only for displaying concerning results (outside the normal range, but inside the critical range) or critical results (outside the critical range, but within the absolute range). Does that make sense? `lowAbsolute` and `hiAbsolute` should be rare and only in cases where the value can never exceed that range (e.g., SpO2 is a percent, so it's always between 0 and 100; heights and weights should never be a negative number, things like that).

ibacher: But we can handle critical icons in a separate PR.

src/hooks/useFormFieldsMeta.ts
```diff
@@ -11,6 +11,20 @@ export function useFormFieldsMeta(rawFormFields: FormField[], concepts: OpenmrsR
         const matchingConcept = findConceptByReference(field.questionOptions.concept, concepts);
         field.questionOptions.concept = matchingConcept ? matchingConcept.uuid : field.questionOptions.concept;
         field.label = field.label ? field.label : matchingConcept?.display;
+        if(matchingConcept)
+        {
+          if(matchingConcept.units)
+          {
+            field.label = field.label + " (" + matchingConcept.units + ")";
```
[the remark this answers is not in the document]
D-matz: @samuelmale thanks, moved the units and range to the number component. Also for a concept that defines min but not max it looks like 
min0

leaving critical alone for now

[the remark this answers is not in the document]
ibacher: We should harmonize the view here to what we're doing [here](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2161), which is `>= 0` or `<= 100` things like that.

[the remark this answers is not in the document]
D-matz: min1

unfortunately it splits the label onto a new line, but the format should match the chart lab result format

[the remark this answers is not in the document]
ibacher: I mean, that was always going to be a possibility... This just made it apparent for that one case.

src/hooks/useFormFieldsMeta.ts
```diff
@@ -11,6 +11,20 @@ export function useFormFieldsMeta(rawFormFields: FormField[], concepts: OpenmrsR
...
+            field.questionOptions.min = matchingConcept.lowAbsolute;
+            field.questionOptions.max = matchingConcept.hiAbsolute;  
+          }
+          else if (matchingConcept.lowAbsolute != undefined) {
+            field.questionOptions.min = matchingConcept.lowAbsolute;
+          }
+          else if (matchingConcept.hiAbsolute != undefined) {
+            field.questionOptions.max = matchingConcept.hiAbsolute;
+          }
+        }
```
samuelmale, line 26:
```suggestion
if (matchingConcept) {
  const { lowAbsolute, hiAbsolute } = matchingConcept;
  
  if (lowAbsolute !== undefined) {
    field.questionOptions.min = lowAbsolute;
  }
  
  if (hiAbsolute !== undefined) {
    field.questionOptions.max = hiAbsolute;
  }
}
```

src/components/inputs/number/number.component.tsx
```diff
@@ -11,6 +11,28 @@ import { useFormProviderContext } from '../../../provider/form-provider';
 import FieldLabel from '../../field-label/field-label.component';
 import { isEmpty } from '../../../validators/form-validator';
 
+
+const extractFieldUnitsAndRange = (concept) => {
```
samuelmale, line 15: Can you wrap this in a [useCallback](https://react.dev/reference/react/useCallback)?

ibacher: No, actually, you shouldn't do that here. This is actually done correctly. Functions defined inside functional components would be swapped with a new instance every time the function is called (so, every time the renderer detects that the component might change). `useCallback()` prevents that by ensuring the variable keeps a stable reference to the same function until the dependency array changes. Since this function isn't defined in a functional component body, references to this function will always be stable references to the same function, so `useCallback()` isn't needed here (and, IMO, we should prefer these ancillary functions be defined outside of functional components wherever feasible).

All that said, the `concept` argument here must have a type defined.

src/components/inputs/number/number.component.tsx
```diff
@@ -61,7 +83,7 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
           id={field.id}
           invalid={errors.length > 0}
           invalidText={errors[0]?.message}
-          label={<FieldLabel field={field} />}
+          label={<FieldLabel field={field} customLabel={t(field.label) + extractFieldUnitsAndRange(field.meta?.concept)} />}
```
ibacher, line 86: Please don't do string concatenation like this with translated strings. Use [i18next's](https://www.i18next.com/translation-function/interpolation) interpolation feature for this.

ibacher: In this case, the code now hard-codes a word order that doesn't make sense in LTR languages.

src/components/inputs/number/number.test.tsx
```diff
@@ -22,6 +22,25 @@ const numberFieldMock = {
   readonly: false,
 };
 
+const numberFieldMockWithUnitsAndRange = {
```
ibacher, line 25: It would be nice to have a more comprehensive set of tests.

src/components/inputs/number/number.component.tsx
```diff
@@ -11,6 +11,28 @@ import { useFormProviderContext } from '../../../provider/form-provider';
...
+  }
+
+  const { hiAbsolute, lowAbsolute, units } = concept;
+  const displayUnit = units ? ` ${units}` : '';
+
+  const hasLowerLimit = lowAbsolute != null;
+  const hasUpperLimit = hiAbsolute != null;
+
+  if (hasLowerLimit && hasUpperLimit) {
+      return ` (${lowAbsolute} - ${hiAbsolute} ${displayUnit})`;
```
ibacher, line 27: Don't hard-code prepended spaces in strings here.

ibacher (changes requested):

src/components/inputs/number/number.component.tsx
D-matz: I changed the label to t('{{fieldDescription}} {{unitsAndRange}}' but not sure if that's correct as I haven't used i18n before or tested a right to left language case. Thanks for the i18n help.

src/components/inputs/number/number.test.tsx
D-matz: added a few more cases (only units, only range, only max)

src/components/inputs/number/number.test.tsx
D-matz, line 1: not really sure about jest.mock('react-i18next', () => ({...
it seemed needed for the new i18n label t('{{fieldDescription}} {{unitsAndRange}}'... but I'm not sure if that's correct

src/hooks/useFormFieldsMeta.ts
```diff
@@ -11,6 +11,20 @@ export function useFormFieldsMeta(rawFormFields: FormField[], concepts: OpenmrsR
         const matchingConcept = findConceptByReference(field.questionOptions.concept, concepts);
         field.questionOptions.concept = matchingConcept ? matchingConcept.uuid : field.questionOptions.concept;
         field.label = field.label ? field.label : matchingConcept?.display;
+        
+        if (matchingConcept) {
+          if (matchingConcept.lowAbsolute != undefined && matchingConcept.hiAbsolute != undefined) {
```
ibacher, line 16: It's always better to use `typeof x !== 'undefined'`, though this kind of works, but only because SWC will re-write the `undefined`s here to `void 0`. In any case, please use exact comparisons (`!==` and `===`).

ibacher: Also, please just implement @samuelmale's suggestion. It's much cleaner.

src/components/inputs/number/number.component.tsx
```diff
@@ -10,6 +10,28 @@ import { useTranslation } from 'react-i18next';
...
+
+const extractFieldUnitsAndRange = (concept?: Concept): string => {
+  if (!concept) {
+    return '';
+  }
+
+  const { hiAbsolute, lowAbsolute, units } = concept;
+  const displayUnits = units ? ` ${units}` : '';  
+  const hasLowerLimit = lowAbsolute != null;
+  const hasUpperLimit = hiAbsolute != null;
```
ibacher, line 24: Again `!==`. But I think here it's preferential to just use `isNil()` from lodash.

src/components/inputs/number/number.component.tsx
```diff
@@ -10,6 +10,28 @@ import { useTranslation } from 'react-i18next';
 import { useFormProviderContext } from '../../../provider/form-provider';
 import FieldLabel from '../../field-label/field-label.component';
 import { isEmpty } from '../../../validators/form-validator';
+import { Concept } from '@openmrs/esm-framework';
```
ibacher, line 13: Please see our [extension documentation](https://o3-docs.openmrs.org/docs/coding-conventions/code-organization) on how imports are to be ordered.

src/components/inputs/number/number.component.tsx
```diff
@@ -61,7 +83,12 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
           id={field.id}
           invalid={errors.length > 0}
           invalidText={errors[0]?.message}
-          label={<FieldLabel field={field} />}
+          label={<FieldLabel field={field} customLabel={t('{{fieldDescription}} {{unitsAndRange}}',
+            {
+              fieldDescription: t(field.label),
+              unitsAndRange: extractFieldUnitsAndRange(field.meta?.concept),
+              interpolation: { escapeValue: false }
```
ibacher, line 90: Why is this necessary?

src/components/inputs/number/number.component.tsx
```diff
@@ -61,7 +83,12 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
           id={field.id}
           invalid={errors.length > 0}
           invalidText={errors[0]?.message}
-          label={<FieldLabel field={field} />}
+          label={<FieldLabel field={field} customLabel={t('{{fieldDescription}} {{unitsAndRange}}',
```
ibacher, line 86: We should have a real key here with this interpolation as the `defaultValue`.

src/components/inputs/number/number.test.tsx
```diff
@@ -3,6 +3,17 @@ import { act, render, screen, fireEvent } from '@testing-library/react';
 import { useFormProviderContext } from 'src/provider/form-provider';
 import NumberField from './number.component';
 
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key, options) => {
+      if (options && 'fieldDescription' in options) {
+        return `${options.fieldDescription} ${options.unitsAndRange || ''}`.trim();
```
ibacher, line 10: Since `extractFieldUnitsAndRange()` always returns a string:

```suggestion
        return `${options.fieldDescription} ${options.unitsAndRange}`.trim();
```

src/components/inputs/number/number.test.tsx
ibacher: Could you add one for lowAbsolute only as well? That should cover things.

ibacher (changes requested):

src/hooks/useFormFieldsMeta.ts
D-matz: applied suggestion @samuelmale ty

src/components/inputs/number/number.component.tsx
D-matz: moved useTranslation up as well

src/components/inputs/number/number.component.tsx
D-matz: without escapeValue: false, / doesn't show correctly, it becomes &#x2F; eg BMI (Kg&#x2F;m2):

src/components/inputs/number/number.component.tsx
D-matz: thanks again for i18n help, I changed the test too to t: (key, defaultValueOrOptions, options), hopefully that's correct now that there's a key and default value but not sure

samuelmale (commented):
@D-matz can you resolve the lint issues and merge conflicts and we have this merged in?

src/components/inputs/number/number.component.tsx
samuelmale: @D-matz did you finally use i18next's interpolation instead of the manual concatenation?

samuelmale: (Just double-checked, marking as resolved)

D-matz:
@samuelmale by the way, I had to add i18n usetranslation to unspecified.test.tsx and form-engine.test.tsx which I am not sure about, for those tests to pass. Also note I have 9 failing tests on yarn test on both this branch and main. So please let me know if I'm doing something wrong with these.

PR #457, opened 2025-01-17
(chore) Transifex migration for translations to @openmrs/esm-form-engine-app

pirupius:
## Requirements

## Summary
As part of the migration to use Transifix and align missing translations from this repo
- Migrated all translation files and deleted any translation setup files 
- Switched translation to use @openmrs/esm-form-engine-app namespace
- Added translation provision for sidenav links

## Screenshots

Sample screenshot in Arabic

Screenshot 2025-01-15 at 10 40 37


src/components/sidebar/sidebar.component.tsx
```diff
@@ -98,7 +99,7 @@ function PageLink({ page, currentActivePage, pagesWithErrors, requestPage }: Pag
           e.preventDefault();
           requestPage(page.id);
         }}>
-        <span>{page.label}</span>
+        <span>{t(page.label)}</span>
```
pirupius, line 102: Adds support for translations to sidenav menu-items

samuelmale:
I hoped this was about setting up Transifex integration for this repo and not migrating the translation setup to the form-engine-app? Could I be missing something @pirupius?

pirupius:
@samuelmale Initially the plan was to integrate Transifex but on second review and interaction with @ibacher, he pointed out that the convention for that integration is for apps and not libraries. That's the same convention that has been implemented for esm-core libraries as well as patient commons lib within the patient chart. 
The focus then changed to moving the translations out of this library and using the `@openmrs/esm-form-engine-app` namespace where the translations will be managed.

samuelmale:
I’m concerned about the tight coupling between the “form-engine-app” and the library. This approach assumes that the form engine is only intended for use within the patient chart. As a result, internationalization (i18n) would break for those using the form engine outside the patient chart, or for users with a custom ESM designed for dispensing forms within the patient chart.

pirupius:
Maybe @vasharma05 and @ibacher can correct me on this;

Based on the setup and the way it has been implemented with other libraries, it isn't necessarily restricting the use of the translations to only the patient chart but it's just using the namespace `@openmrs/esm-form-engine-app` to pick the translations. Similarly to the way many custom esms have the `esm-commons-lib` as a dependency but the translations are picked from the `esm-patient-chart`.


This is an example with form builder with the local version of my `form-engine-app` linked with the translations added in this [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2200). You can see the `Save` and `Cancel` translations which were previously in the `form-engine-lib` are being picked automatically from the loaded version of the `form-engine-app`

Screenshot 2025-01-17 at 13 52 02

The catch would be if some is using the library but not using the form-engine-app which sounds unlikely.

denniskigen:
Could we ad something to the README explaining how these translations are wired up following this change so folks used to the previous approach don't get tripped up?

denniskigen:
@samuelmale Tangential to these changes, but [ErrorFallback](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/components/renderer/field/form-field-renderer.component.tsx#L218 should not use the `ToastNotification` component - it should use a Snackbar instead in-line with our [notification design guidelines](https://zeroheight.com/23a080e38/p/683580-notifications).

denniskigen (approved):
LGTM. See comment about documenting the new i18n approach for posterity.

ibacher:
Just a note on the coupling here: because of [how we load translations](https://github.com/openmrs/openmrs-esm-core/blob/000976675fa24a2f6b442b3a29bfca5b29187f59/packages/shell/esm-app-shell/src/locale.ts#L34), we kind of have the option of either:

1. Coupling the form engine lib to the form engine app for the source of the translations
2. Coupling core itself to the form engine lib to do something like the work-around we have to load "core" translations
3. Modify the form engine lib to "push" it's translations into the i18next cache

1 is, as @pirupius mentions, something we're already doing and seemed clearly superior to 2. 

3 only just occurred to me as I was writing this up and might be logically cleaner, but I think there are some technical challenges with being able to accommodate language changes. Roughly option 3 looks like:

In `I18FormEngine`:
1.  In an effect hook, call `i18next.addResourceBundle()` with an appropriate namespace.
2. In another effect hook, register a listener for the `languageChanged` event. When this event fires, retrigger 1.

samuelmale:
I'm inclined to the third option. Let's see how this works out! In case of any backlash inspired by the coupling problem, we should probably spike at the third option.

PR #464, opened 2025-02-09
Adding conformation modal for empty form

UNCANNY69:
## Requirements

## Summary

This PR adds a feature which triggers a conformation modal when User submits an empty form.

## Screenshots

https://github.com/user-attachments/assets/ae95f747-dcf1-41c2-b06d-5f42a071a814

### DIFF
--- a/src/empty-form-conformation-modal.tsx
+++ b/src/empty-form-conformation-modal.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import { ComposedModal, ModalHeader, ModalBody, ModalFooter, Button } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
+import styles from './empty-formconformation-modal.scss'
+
+interface EmptyFormConfirmationModalProps {
+  onDiscard: () => void;
+  onConfirmation: () => void;
+  open: boolean;
+}
+
+const IncompleteFormConfirmationModal: React.FC<EmptyFormConfirmationModalProps> = ({
+  onDiscard,
+  onConfirmation,
+  open,
+}) => {
+  const { t } = useTranslation();
+
+  return (
+    <ComposedModal open={open} onClose={onDiscard} className={styles.customModal}  size="small">
+      <ModalHeader title={t('EmptyForm', 'Empty Form ')}  />
+      <ModalBody>
+        <p >
+          {t(
+            'EmptyFormConfirmation',
+            'All fields are Empty. Are you sure you want to submit the form?',
+          )}
+        </p>
+      </ModalBody>
+      <ModalFooter >
+        <Button kind="secondary" onClick={onDiscard}>
+          {t('cancel', 'Cancel')}
+        </Button>
+        <Button kind="danger" onClick={onConfirmation}>
+          {t('confirm', 'Confirm')}
+        </Button>
+      </ModalFooter>
+    </ComposedModal>
+  );
+};
+
+export default IncompleteFormConfirmationModal;
--- a/src/empty-formconformation-modal.scss
+++ b/src/empty-formconformation-modal.scss
@@ -1,0 +1,13 @@
+@use '@carbon/react/scss/colors';
+
+
+.customModal  {
+    :global(.cds--modal-close) {
+        position: fixed;
+      }
+
+    :global( .cds--btn--primary) {
+        background-color: #f4f4f4;
+      }
+     
+}   
--- a/src/provider/form-factory-helper.ts
+++ b/src/provider/form-factory-helper.ts
@@ -5,6 +5,7 @@ import { evaluatePostSubmissionExpression } from '../utils/post-submission-actio
 import { type PostSubmissionActionMeta } from '../hooks/usePostSubmissionActions';
 import { type TFunction } from 'react-i18next';
 import { type SessionMode } from '../types';
+import { Console } from 'console';
 
 export function validateForm(context: FormContextProps) {
   const {
@@ -44,6 +45,24 @@ export function validateForm(context: FormContextProps) {
     )
     .filter((error) => Boolean(error));
   return errors.length === 0;
+}
+
+export function validateEmptyFields(context: FormContextProps){
+  const {
+    formFields,
+    formFieldValidators,
+    patient,
+    sessionMode,
+    addInvalidField,
+    updateFormField,
+    methods: { getValues, trigger },
+  } = context;
+  const values = getValues();
+  return Object.values(values).every(value => 
+    value === null || 
+    value === undefined || 
+    (Array.isArray(value) && value.length === 0)
+  );
 }
 
 export async function processPostSubmissionActions(
--- a/src/provider/form-factory-provider.tsx
+++ b/src/provider/form-factory-provider.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useCallback, useContext, useEffect, useRef } from 'react';
+import React, { createContext, useCallback, useContext, useEffect, useRef, useState } from 'react';
 import { type FormField, type FormSchema, type SessionMode } from '../types';
 import { EncounterFormProcessor } from '../processors/encounter/encounter-form-processor';
 import {
@@ -11,9 +11,10 @@ import {
 } from '@openmrs/esm-framework';
 import { type FormProcessorConstructor } from '../processors/form-processor';
 import { type FormContextProps } from './form-provider';
-import { processPostSubmissionActions, validateForm } from './form-factory-helper';
+import { processPostSubmissionActions, validateForm, validateEmptyFields } from './form-factory-helper';
 import { useTranslation } from 'react-i18next';
 import { usePostSubmissionActions } from '../hooks/usePostSubmissionActions';
+import IncompleteFormConfirmationModal from '../empty-form-conformation-modal';
 
 interface FormFactoryProviderContextProps {
   patient: fhir.Patient;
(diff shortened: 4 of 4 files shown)


UNCANNY69:
I made many changes to the form-factory-provider , I would like it to be reviewed in case it can be designed better @NethmiRodrigo @samuelmale @denniskigen

src/empty-form-conformation-modal.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useTranslation } from 'react-i18next';
+import styles from './empty-formconformation-modal.scss'
+
+interface EmptyFormConfirmationModalProps {
+  onDiscard: () => void;
+  onConfirmation: () => void;
+  open: boolean;
+}
+
+const IncompleteFormConfirmationModal: React.FC<EmptyFormConfirmationModalProps> = ({
```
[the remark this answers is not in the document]
NethmiRodrigo: For more context @UNCANNY69, for modals the standard is to use the modal system where we don't need to have states to trigger modal rendering, you can read more about it in the docs- https://o3-docs.openmrs.org/docs/modal-system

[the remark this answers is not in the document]
NethmiRodrigo: For more context, the standard is to use the modal system - https://o3-docs.openmrs.org/docs/modal-system, where we don't have to depend on a state to trigger a modal

src/empty-form-conformation-modal.tsx
denniskigen, line 1: Typo in the filename - it should be [something like](https://o3-docs.openmrs.org/docs/coding-conventions/naming) `empty-form-confirmation.modal.tsx`

src/empty-formconformation-modal.scss
denniskigen, line 1: Please run the linter on this file. The spacing an indentation look off. Also, the filename should be something like `empty-form-confirmation.modal.scss`.

src/empty-form-conformation-modal.tsx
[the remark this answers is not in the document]
denniskigen: Sounds right to me. And IIRC, this modal likely needs to exist and be registered in the [Form Engine app](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-form-engine-app/src/routes.json#L18) in Patient Chart, not in the library itself. Is that right, @samuelmale?

UNCANNY69:
Reated PR:

https://github.com/openmrs/openmrs-esm-patient-chart/pull/2237

UNCANNY69:
I made the changes according to the discussion in the JIRA, can it be reviewed again @denniskigen @samuelmale @NethmiRodrigo

src/provider/form-factory-provider.tsx
```diff
@@ -96,14 +100,31 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
-      if (isValid) {
-        Promise.all(forms.map((formContext) => formContext.processor.processSubmission(formContext, abortController)))
-          .then(async (results) => {
+    const handleFormSubmission = async () => {
+      if (isSubmitting) {
+        const forms = [rootForm.current, ...Object.values(subForms.current)];
+        // Check if the form is empty
+        const isEmpty = forms.every((formContext) => validateEmptyForm(formContext));
+        // Validate all forms
+        const isValid = forms.every((formContext) => validateForm(formContext));
```
samuelmale, line 109: This validation should happen before the "Empty" validation.

src/provider/form-factory-provider.tsx
```diff
@@ -96,14 +100,31 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
+
+        if (isEmpty) {
```
samuelmale, line 111: We want to do something like:

```ts
// If the form is valid but empty, prompt user 
if (isValid && isEmpty) {
   // prompt user ...
}

```

src/form-engine.component.tsx
```diff
@@ -127,6 +129,8 @@ const FormEngine = ({
           provider={session?.currentProvider}
           visit={visit}
           handleConfirmQuestionDeletion={handleConfirmQuestionDeletion}
+          handleEmptyFormSubmission={handleEmptyFormSubmission}
+          handleDiscardForm={handleClose}
```
samuelmale, line 133: You may not have to add this considering that `handleClose` is basically a workspace property.

samuelmale: After a second look, I realized you added this property to the `FormFactoryProvider` as opposed to the root component as I had previously assumed, my bad! It seems like we are lifting the `handleClose` property twice, any reason for not using `formSubmissionProps.handleClose`?

UNCANNY69: Cause that handleClose is passed as a void function in the formProps

UNCANNY69: image

samuelmale: Ooppss!! That's something we should fix. Can you update those props with valid values from the parent component and use that instead? (It's probably safe to get rid of the `onError` prop)

UNCANNY69: There was an issue when i updated it with the value the form never opened ig that's the reason it was not passed as for removing onError properly i can do it.

samuelmale: Which error do you get exactly?

UNCANNY69: I resolved it and utilizes only handleClose @samuelmale , can it be reviewed again ?

src/form-engine.component.tsx
```diff
@@ -127,13 +129,13 @@ const FormEngine = ({
...
+          handleEmptyFormSubmission={handleEmptyFormSubmission}
           isFormExpanded={isFormExpanded}
           formSubmissionProps={{
             isSubmitting,
             setIsSubmitting,
-            onSubmit,
-            onError: () => {},
-            handleClose: () => {},
+            onSubmit, 
+            handleClose: handleClose,
```
vasharma05, line 138: Why is onError removed from here?

src/provider/form-factory-helper.ts
```diff
@@ -46,6 +46,24 @@ export function validateForm(context: FormContextProps) {
...
+    updateFormField,
+    methods: { getValues, trigger },
+  } = context;
+  const values = getValues();
+  return Object.values(values).every(value => 
+    value === null || 
+    value === undefined || 
+    (Array.isArray(value) && value.length === 0)
+  );
+}
```
vasharma05, line 65: A better way for this is to use the `methods.isDirty` value from the `methods` on line 57.

src/form-engine.component.tsx
UNCANNY69: It was not being used and @samuelmale suggested in the above comments to remove the attribute, should i add it back ?

src/provider/form-factory-helper.ts
UNCANNY69: Yes @vasharma05 , it makes it much more simpler to use the dirtyFields , makes performance much faster

src/provider/form-factory-provider.tsx
```diff
@@ -96,14 +97,31 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
+        const forms = [rootForm.current, ...Object.values(subForms.current)];
+        // Validate all forms
+        const isValid = forms.every((formContext) => validateForm(formContext));
+        // Check if the form is empty
+        const isEmpty = forms.every((formContext) => validateEmptyForm(formContext));
+
+        if (isEmpty && isValid) {
+          if (handleEmptyFormSubmission && typeof handleEmptyFormSubmission === 'function') {
+            try {
+              await handleEmptyFormSubmission();
```
samuelmale, line 111: This should basically launch a modal prompting the user. Based on the user's decision, we should either save an empty encounter by submitting the form or close the form. Here is an [example](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/734f0e590ea359f883c8bde68b72d15b9cac260c/src/components/repeat/repeat.component.tsx#L123-L136).

UNCANNY69: I think in the current scenario the user is not allowed to submit empty forms, thus there is no saving the form.

src/provider/form-factory-helper.ts
```diff
@@ -46,6 +46,11 @@ export function validateForm(context: FormContextProps) {
   return errors.length === 0;
 }
 
+export function validateEmptyForm(context: FormContextProps) {
+  const { methods: { formState: { dirtyFields } } } = context;
+  return Object.keys(dirtyFields).length === 0;
```
samuelmale, line 51: Can you filter out [transient](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/e92c22c63f37c8b21cd5b97c39cef5b8d44dab31/src/types/schema.ts#L154) fields?

UNCANNY69:
Resolved all the comments can it be reviewed again @denniskigen @samuelmale ?

src/provider/form-factory-provider.tsx
```diff
@@ -96,14 +97,42 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
+  
+        if (isValid) {
+          // Check if the form is empty
+          const isEmpty = forms.every((formContext) => validateEmptyForm(formContext));
+          if (isEmpty) {
+            if (handleEmptyFormSubmission && typeof handleEmptyFormSubmission === 'function') {
+              const result = handleEmptyFormSubmission(); 
+              if (result instanceof Promise) { 
+                try {
+                  await result;
```
samuelmale, line 114: Shouldn't we try inferring the resolved promise to a boolean?

PR #492, opened 2025-04-17
(fix) Undefined `min` and `max` values for the number inputs

vasharma05:
## Requirements

## Summary
The Carbon Input by default restricts the value change by the steppers between min and max values. But this restriction isn't working for the form engine. I went to the [Carbon NumberInput code for the min and max restriction](https://github.com/carbon-design-system/carbon/blob/0ce91ed5ad791ce04fa7878f2d3cffe7496267c3/packages/react/src/components/NumberInput/NumberInput.tsx#L376) and realized that the min and max value we might be sending is actually `undefined`, and if any value among the min/max values is `undefined`, the restriction doesn't work.

I went to the code and realized that the final value for the min and max `value` is: `Number(field.questionOptions.min) || undefined`. Now, if the `field.questionOptions.min = 0`, the value passed isn't 0, but `undefined`. The `||` check must be replaced by `??`. I have added a better check for getting the correct value.

### Before

https://github.com/user-attachments/assets/3ffec393-d7b3-4750-8880-f13685b9237b

### After

https://github.com/user-attachments/assets/908ed244-0329-4e24-a5c2-2f04d1b80ab0

### DIFF
--- a/src/components/inputs/number/number.component.tsx
+++ b/src/components/inputs/number/number.component.tsx
@@ -22,12 +22,17 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
     return value ?? '';
   }, [value]);
 
+  const getNumericValue = useCallback(
+    (value: string | number) => (typeof value === 'undefined' || isNaN(Number(value)) ? undefined : Number(value)),
+    [],
+  );
+
   const handleChange = useCallback(
     (event, { value }) => {
-      const parsedValue = isEmpty(value) ? undefined : Number(value);
+      const parsedValue = getNumericValue(value);
       setFieldValue(isNaN(parsedValue) ? undefined : parsedValue);
     },
-    [setFieldValue],
+    [setFieldValue, getNumericValue],
   );
 
   const isInline = useMemo(() => {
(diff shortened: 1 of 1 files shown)


src/components/inputs/number/number.component.tsx
```diff
@@ -22,12 +22,17 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
...
 
+  const getNumericValue = useCallback(
+    (value: string | number) => (typeof value === 'undefined' || isNaN(Number(value)) ? undefined : Number(value)),
+    [],
+  );
+
   const handleChange = useCallback(
     (event, { value }) => {
-      const parsedValue = isEmpty(value) ? undefined : Number(value);
+      const parsedValue = getNumericValue(value);
```
pirupius, line 32: Does this alter the value on submission? for example if i don't enter any value in the number field, does the value ignored because it's undefined or is this parsed as zero?

samuelmale:
@vasharma05, the attached videos seem to be corrupted or something?

src/components/inputs/number/number.component.tsx
vasharma05: If there's nothing in the input field, the `value` returned by the `onChange` is `''`, hence marking it as undefined rather than `''`.

vasharma05:
> @vasharma05, the attached videos seem to be corrupted or something?

Why so @samuelmale?

src/components/inputs/number/number.component.tsx
```diff
@@ -37,6 +42,9 @@ const NumberField: React.FC<FormFieldInputProps> = ({ field, value, errors, warn
     return false;
   }, [sessionMode, field.readonly, field.inlineRendering, layoutType, workspaceLayout]);
 
+  const max = getNumericValue(field.questionOptions.max);
+  const min = getNumericValue(field.questionOptions.min);
```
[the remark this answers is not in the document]
vasharma05: I see no harm in this way, it's more explicit that these exact values are passed to the component.

pirupius (approved):

PR #539, opened 2025-05-16
(chore) Adjust update-openms-deps.yml to run once a week

pirupius:
## Requirements

## Summary
This cron has been running every 30mins and has been adding a bunch of commits even if nothing much has been changing on the FE lib side of things. This change adjusts it to run once a week (Monday at midnight)

### DIFF
--- a/.github/workflows/update-openmrs-deps.yml
+++ b/.github/workflows/update-openmrs-deps.yml
@@ -2,7 +2,7 @@ on:
   workflow_dispatch:
   schedule:
     # every hour
-    - cron: '30 * * * *'
+    - cron: '30 0 * * 1'
 
 name: 'Check for OpenMRS Dependency Updates'
 
--- a/.github/workflows/update-openmrs-deps.yml
+++ b/.github/workflows/update-openmrs-deps.yml
@@ -1,7 +1,7 @@
 on:
   workflow_dispatch:
   schedule:
-    # every hour
+    # Once every week on Monday at 00:30
     - cron: '30 0 * * 1'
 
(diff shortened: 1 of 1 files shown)


samuelmale:
@pirupius, What do you think about having the job run once a day?

.github/workflows/update-openmrs-deps.yml
```diff
@@ -2,7 +2,7 @@ on:
   workflow_dispatch:
   schedule:
     # every hour
```
samuelmale, line 4: You may have to update the comment.

pirupius:
@samuelmale the reason i put it to one week was because the FE lib isn't updated as much as the patient chart or patient management. Case in point is that the last bug fix/ feature was merged in 20 days ago and since then it's been mostly the automated bump openmrs dependencies. In my view it looked like once a week is more reasonable.

ibacher:
The point of this workflow, though, is to update dependencies that this depends on, including the framework. It only generates PRs if they are actually changed. 

I'm ok with ramping it down, but it's important to keep in mind what this is actually for.

samuelmale:
> Once a week is way too long between things if updating versions introduces bugs.

@ibacher, are you fine with triggering the job once a day?

ibacher:
> are you fine with triggering the job once a day?

Yes, but I'm really not understanding what the issue is that this is fixing.

FWIW, as-is, this job runs hourly at 30 minutes past the hour. The crontab for "every 30 minutes" is something like `*/30 * * * * *`.

ibacher:
I moved the job to be daily.

PR #550, opened 2025-06-03
(feat): Enable external form Submission

Twiineenock:
## Requirements

## Summary
This PR introduces two custom React hooks — `useExternalSubmitListener` and `usePostSubmissionCallback` — into the FormEngine to improve its interoperability with external shells like the Form Display Engine (FDE).

#### 1. `useExternalSubmitListener`

Enables **external triggering** of form submission by listening for a global custom event (`triger-form-engine-submit`). This is especially useful when the FormEngine is embedded in another module's UI like the FDE, and submission needs to be controlled from outside (e.g., a shell toolbar or modal footer).

* Registers a `window` event listener on mount.
* Invokes the provided `submitFn` when the event is dispatched.
* Cleans up the listener on unmount.

#### 2. `usePostSubmissionCallback`

Supports **deferring follow-up actions** until after form submission completes. This allows external code to safely run logic (e.g., closing modals, navigation, showing notifications) *after* the form has finished submitting.

* Listens for a `form-submission-complete` custom event.
* Executes an action immediately if the form has already submitted, or defers it until `setIsFormSubmitted(true)` is called.
* Prevents duplicate executions.

### 🧩 Use Case: Form Display Engine (FDE)

These hooks make it possible to:

* Trigger form submission from an external shell.
* Respond to submission completion in a decoupled, reliable way.

This integration improves reusability of the `FormEngine` across platforms while preserving single-responsibility and encapsulation within the component.

### 🧪 How they are used

1. Use `useExternalSubmitListener` to listen for submission trigger events.
2. Trigger a form submission by dispatching:

   ```ts
   window.dispatchEvent(new Event('triger-form-engine-submit'));
   ```
3. Optionally, dispatch a post-submission callback:

   ```ts
   window.dispatchEvent(
     new CustomEvent('form-submission-complete', {
       detail: { action: () => console.log('Post-submit logic ran!') }
     })
   );
   ```

### ✅ Checklist

* [x] Custom hooks added with full documentation
* [x] Submission now triggerable externally
* [x] One-time post-submission callbacks supported
* [x] Backwards compatibility maintained
* [x] PR reviewed and tested locally

## Screenshots

RFE-4-FDE

## Screencasts
https://www.loom.com/share/193799987e0848a0b20e0575a462a3de?sid=c0d5143c-7aac-4e5f-bedd-aa3671833ffe

## Other

Fast Data Entry: [https://github.com/openmrs/openmrs-esm-fast-data-entry-app/pull/115](https://github.com/openmrs/openmrs-esm-fast-data-entry-app/pull/115)

Patient Chart: [https://github.com/openmrs/openmrs-esm-patient-chart/pull/2476](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2476)


src/hooks/usePostSubmissionCallback.ts
```diff
@@ -0,0 +1,74 @@
...
+ * setIsFormSubmitted(true);
+ *
+ * // Somewhere else in the app, dispatch a custom event:
+ * window.dispatchEvent(
+ *   new CustomEvent('form-submission-complete', {
+ *     detail: { action: () => console.log('Post-submission logic executed') }
+ *   })
+ * );
+ */
+export function usePostSubmissionCallback() {
```
samuelmale, line 36: Don't we achieve the same utility from [onSubmit](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/form-engine.component.tsx#L31)?

src/hooks/useExternalSubmitListener.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param eventName - (Optional) The name of the custom event to listen for. Defaults to `'triger-form-engine-submit'`.
+ *
+ * @example
+ * // Inside your form component
+ * useExternalSubmitListener(() => handleSubmit());
+ *
+ * // Somewhere else (e.g., external button or shell app)
+ * window.dispatchEvent(new Event('triger-form-engine-submit'));
+ */
+export function useExternalSubmitListener(submitFn: () => void, eventName = 'triger-form-engine-submit') {
```
samuelmale, line 26: If we are to make this hook more specific, I would change its signature slightly:

```suggestion
export function useExternalSubmitListener(internalSubmitHandler: () => void) {
```
If the hook is defining a specific listener, I suggest defining the `eventName` locally and renaming it to something like "rfe-form-submit-action" or "rfe-trigger-form-submit".

src/hooks/useExternalSubmitListener.ts
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    const handleSubmitWrapper = () => submitFn();
+    window.addEventListener(eventName, handleSubmitWrapper);
```
samuelmale, line 29: We should expect the event to provide more details about the target form session ie. provide the target form and patient UUID. This would help us map the dispatched event to the correct form instance: 

```ts
const { formUuid, patientUuid } = useFormContext();
const handleSubmit = (event) => {
    const { formUuid: targetForm, patientUuid: targetPatient, encounterUuid /**while in edit mode **/} = event.details;
    if (targetForm === formUuid && targetPatient === patientUuid) {
         // invoke submission
         internalSubmitHandler();
    }
}

```

src/components/sidebar/sidebar.component.tsx
```diff
@@ -16,6 +16,7 @@ interface SidebarProps {
   onCancel: () => void;
   handleClose: () => void;
   hideFormCollapseToggle: () => void;
+  isSubmissionTriggeredExternally?: boolean;
```
samuelmale, line 19: I would name this something like `hideControls` or `hideInternalButtons` which defaults to `false`.

samuelmale (changes requested):
@Twiineenock do we also need to support an event that triggers validation?

Twiineenock:
> @Twiineenock do we also need to support an event that triggers validation?

> @samuelmale Good question!
> So far, I haven't encountered a need for a separate validation trigger. The RFE automatically performs validation beforee it can submit a form.

src/hooks/useExternalSubmitListener.ts
Twiineenock: Updated

src/hooks/useExternalSubmitListener.ts
Twiineenock: updated

Twiineenock:
I have updated the rest of the PRs within other repos to match these new changes

src/hooks/useExternalSubmitListener.ts
```diff
@@ -0,0 +1,57 @@
+import { useEffect } from 'react';
+import { createGlobalStore, getGlobalStore, useStore } from '@openmrs/esm-framework';
+
+interface SubmitEventDetail {
+  formUuid: string;
+  patientUuid: string;
+  action: string;
```
samuelmale, line 7: Since our hook isn't generic, let's get rid of the action prop.

src/hooks/useExternalSubmitListener.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type InternalSubmitHandler = () => void;
+
+//State that holds the current formUuid and patientUuid.
+createGlobalStore<Record<FormSession, string>>('rfe-FormSession', {
+  formUuid: '',
+  patientUuid: '',
+});
+
+const formSessionStore = getGlobalStore('rfe-FormSession');
```
samuelmale, line 20: This seems overkill! Just consume the props from the root component.

src/form-engine.component.tsx
```diff
@@ -33,6 +34,8 @@ interface FormEngineProps {
   handleClose?: () => void;
   handleConfirmQuestionDeletion?: (question: Readonly<FormField>) => Promise<void>;
   markFormAsDirty?: (isDirty: boolean) => void;
+  hideControls?: boolean;
+  preFilledQuestions?: Record<string, string>;
```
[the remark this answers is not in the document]
Twiineenock: Yes, @samuelmale. The Angular Form Engine (AFE) implements comparable functionality in its form creation service. You can see the relevant logic [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/7df0716f0f88967f50ed61aee129ffdba3be802c/packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts#L284).

src/transformers/default-schema-transformer.ts
```diff
@@ -308,3 +311,23 @@ function handleDiagnosis(question: FormField) {
     delete question.questionOptions['dataSource'];
   }
 }
+
+function handlePreFilledQuestions(form: FormSchema, preFilledQuestions: PreFilledQuestions) {
+  Object.entries(preFilledQuestions).forEach(([prefilledQnId, prefilledValue]) => {
```
samuelmale, line 316: Could you be more consistent with your use of 'pre-filled' vs 'prefilled'?

src/types/index.ts
```diff
@@ -146,5 +147,7 @@ export interface ValidationResult {
   message: string;
 }
 
+export type PreFilledQuestions = Record<string, string>;
```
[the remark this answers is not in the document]
Twiineenock: Thanks for the review, @samuelmale!

Looking at the typing used in both [[form-entry-app](https://github.com/openmrs/openmrs-esm-patient-chart/blob/5bf24de8e33826905048bd0fc4fa0774353e43be/packages/esm-form-entry-app/src/app/types/index.ts#L160)](https://github.com/openmrs/openmrs-esm-patient-chart/blob/5bf24de8e33826905048bd0fc4fa0774353e43be/packages/esm-form-entry-app/src/app/types/index.ts#L160) and [[fast-data-entry](https://github.com/openmrs/openmrs-esm-fast-data-entry-app/blob/7d83c241ec7a8507197bee16824f000efa864c1f/src/FormBootstrap.tsx#L103)](https://github.com/openmrs/openmrs-esm-fast-data-entry-app/blob/7d83c241ec7a8507197bee16824f000efa864c1f/src/FormBootstrap.tsx#L103), we’re following a similar pattern — though with slightly cleaner and more idiomatic syntax in this case.

From the Fast Data Entry side, we typically receive string values (often concept UUIDs) as resolved answers. At this point, FDE isn’t concerned with how different form engines interpret rendering types — it simply passes along the pre-filled values.

The engines themselves are responsible for interpreting these values appropriately:

* Setting them as defaults for their respective questions.
* Letting the field components cast them into the correct data types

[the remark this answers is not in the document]
Twiineenock: Got it @samuelmale! Fixed it!

src/hooks/useExternalSubmitListener.ts
```diff
@@ -0,0 +1,61 @@
...
+ *     ...
+ *   </form>
+ * );
+ */
+
+export function useExternalSubmitListener({ formRef, patientUuid, formUuid }: UseExternalSubmitListenerProps) {
+  useEffect(() => {
+    const handleSubmit = (event: Event) => {
+      const customEvent = event as CustomEvent<SubmitEventDetail>;
+      const { formUuid: targetFormUuid, patientUuid: targetPatientUuid } = customEvent.detail;
```
samuelmale, line 51: @Twiineenock, can you log some kind of error if the form details, ie, form UUID and patient UUID, were not provided?

src/hooks/useExternalSubmitListener.ts
```diff
@@ -47,12 +47,30 @@ interface UseExternalSubmitListenerProps {
...
+          "Event detail is missing 'formUuid' or 'patientUuid'. Both are required to identify and submit the correct form instance.",
+        );
+      }
+
+      if (targetFormUuid === formUuid && targetPatientUuid === patientUuid) {
         formRef.current?.requestSubmit?.();
+      } else {
+        throw new Error(
+          "The provided 'formUuid' or 'patientUuid' in the event detail does not match the expected values for the current form instance.",
+        );
```
samuelmale, line 70: Can you remove this? Let's silently ignore...

src/hooks/useExternalSubmitListener.ts
```diff
@@ -47,12 +47,30 @@ interface UseExternalSubmitListenerProps {
 export function useExternalSubmitListener({ formRef, patientUuid, formUuid }: UseExternalSubmitListenerProps) {
   useEffect(() => {
     const handleSubmit = (event: Event) => {
+      if (!formRef?.current) {
+        throw new Error(
+          "Form reference is missing or not attached. Ensure the 'formRef' is correctly passed and the form element is mounted.",
+        );
```
samuelmale, line 53: Since the user doesn't have much control over this, I would simply log an error as opposed to throwing.  

```ts
console.error("Something went wrong: form reference is missing or not attached. Ensure 'formRef' is correctly passed and the form element is mounted.");
```

ibacher: Ideally, though, the user should get some notification that the task failed. I think we may have a notification system for something like that?

samuelmale: We have a utility for reporting errors [here](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/443696ed43de6ab2297c9772e4ec4573ddc3a9b1/src/utils/error-utils.ts#L3).

samuelmale:
>I.e., in the handler, we should only actually submit the form if the submission event matches the form in this instance.

@ibacher that is being handled [here](https://github.com/Twiineenock/openmrs-esm-form-engine-lib/blob/73be419a35d66a46bbb0d3386928739d66fa577f/src/hooks/useExternalSubmitListener.ts#L65-L67).

src/provider/form-factory-provider.tsx
```diff
@@ -95,12 +99,32 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
     EncounterFormProcessor: EncounterFormProcessor,
   });
 
+  const vidateAllForms = () => {
```
samuelmale, line 102: Two concerns:

1. You've a typo in the function name, so rename it to `validateAllForms`
2. Memoize the function by using a `useCallback(...)`

src/provider/form-factory-provider.tsx
```diff
@@ -95,12 +99,32 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
+    patientUuid: patientUUID,
+    formUuid: formJson?.uuid,
+    setIsSubmitting: setIsSubmitting,
+    setIsValidating: setIsValidating,
+  });
+
+  useEffect(() => {
+    if (isValidating) vidateAllForms();
+    setIsValidating(false);
+  }, [isValidating, setIsValidating, vidateAllForms]);
```
samuelmale, line 121:
```suggestion
  useEffect(() => {
    if (isValidating) {
      validateAllForms();
      setIsValidating(false);
    }
  }, [isValidating, validateAllForms]);
```

src/provider/form-factory-provider.tsx
```diff
@@ -95,12 +99,32 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
+
   useEffect(() => {
     if (isSubmitting) {
       // TODO: find a dynamic way of managing the form processing order
-      const forms = [rootForm.current, ...Object.values(subForms.current)];
       // validate all forms
-      const isValid = forms.every((formContext) => validateForm(formContext));
+      const { forms, isValid } = vidateAllForms();
```
samuelmale, line 127: You may also have to add `validateAllForms` as a hook dependency.

Twiineenock: Truly, have addressed these. Thanks!

src/provider/form-factory-provider.tsx
```diff
@@ -95,12 +99,34 @@ export const FormFactoryProvider: React.FC<FormFactoryProviderProps> = ({
...
   });
 
+  const validateAllForms = useCallback(() => {
+    const forms = [rootForm.current, ...Object.values(subForms.current)];
+    const isValid = forms.every((formContext) => validateForm(formContext));
+    return {
+      forms: forms,
+      isValid: isValid,
+    };
+  }, [rootForm, subForms, validateForm]);
```
samuelmale, line 109:
```suggestion
  }, []);
```

src/types/index.ts
```diff
@@ -146,5 +147,7 @@ export interface ValidationResult {
   message: string;
 }
 
+export type PreFilledQuestions = Record<string, string | number | Date | Array<string | number>>;
```
samuelmale, line 150:
```suggestion
export type PreFilledQuestions = Record<string, string | number | Date | boolean | Array<string>>;
```

samuelmale (commented):
A few minor suggestions.

samuelmale (approved):
LGTM

PR #634, opened 2025-08-27
(fix): Add fallback for hideUnansweredQuestionsInReadonlyForms config

NethmiRodrigo:
## Requirements

## Summary

This PR adds a try-catch block around the external module configuration retrieval to prevent the component from crashing when the configuration from `@openmrs/esm-form-engine-app` module cannot be loaded. This is needed for cases where like the form-builder where the `@openmrs/esm-form-engine-app` is not used.  The component now gracefully falls back to a default value (false) for hideUnansweredQuestionsInReadonlyForms and logs a warning when configuration loading fails.

## Screenshots

After:

https://github.com/user-attachments/assets/cafbf652-df19-4b95-919c-3c190d7a6a3d

Before:

https://github.com/user-attachments/assets/41a07bbc-7df4-4dc9-b067-b7f1b1c45012

### DIFF
--- a/src/components/renderer/field/form-field-renderer.component.tsx
+++ b/src/components/renderer/field/form-field-renderer.component.tsx
@@ -53,7 +53,6 @@ export const FormFieldRenderer = ({ fieldId, valueAdapter, repeatOptions }: Form
       'Could not load external module config, using default value for hideUnansweredQuestionsInReadonlyForms:',
       error,
     );
-    hideUnansweredQuestionsInReadonlyForms = false;
   }
 
   const {
--- a/src/components/renderer/field/form-field-renderer.component.tsx
+++ b/src/components/renderer/field/form-field-renderer.component.tsx
@@ -50,7 +50,7 @@ export const FormFieldRenderer = ({ fieldId, valueAdapter, repeatOptions }: Form
   } catch (error) {
     // If external module config is not available, use default value
     console.warn(
-      'Failed to load @openmrs/esm-form engine-app config - using hideUnansweredQuestionsInReadonlyForms=false (empty fields will be visible in readonly mode): '
+      'Failed to load @openmrs/esm-form engine-app config - using hideUnansweredQuestionsInReadonlyForms=false (empty fields will be visible in readonly mode): ',
       error,
     );
   }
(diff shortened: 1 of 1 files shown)


src/components/renderer/field/form-field-renderer.component.tsx
```diff
@@ -39,9 +39,22 @@ export const FormFieldRenderer = ({ fieldId, valueAdapter, repeatOptions }: Form
...
+      externalModuleName: '@openmrs/esm-form-engine-app',
+    });
+    hideUnansweredQuestionsInReadonlyForms = config?.hideUnansweredQuestionsInReadonlyForms ?? false;
+  } catch (error) {
+    // If external module config is not available, use default value
+    console.warn(
+      'Could not load external module config, using default value for hideUnansweredQuestionsInReadonlyForms:',
+      error,
+    );
+    hideUnansweredQuestionsInReadonlyForms = false;
```
samuelmale, line 56: At this point, I think this reassignment may not be necessary.

src/components/renderer/field/form-field-renderer.component.tsx
```diff
@@ -39,9 +39,22 @@ export const FormFieldRenderer = ({ fieldId, valueAdapter, repeatOptions }: Form
...
+  let hideUnansweredQuestionsInReadonlyForms = false;
+  try {
+    const config = useConfig({
+      externalModuleName: '@openmrs/esm-form-engine-app',
+    });
+    hideUnansweredQuestionsInReadonlyForms = config?.hideUnansweredQuestionsInReadonlyForms ?? false;
+  } catch (error) {
+    // If external module config is not available, use default value
+    console.warn(
+      'Could not load external module config, using default value for hideUnansweredQuestionsInReadonlyForms:',
```
denniskigen, line 53:
```suggestion
      'Failed to load @openmrs/esm-form engine-app config - using hideUnansweredQuestionsInReadonlyForms=false (empty fields will be visible in readonly mode): '
```

Might be better to explain the practical effect that failing to load the module config has.

src/components/renderer/field/form-field-renderer.component.tsx
denniskigen: Yes, `hideUnansweredQuestionsInReadonlyForms` is already `false`. If an error occurs, the try block fails and the assignment on line 49 is never executed. When the catch block runs, the variable remains `false`, so setting it to `false` again has no effect.

NethmiRodrigo: Yeah, that was a left over, my bad. Fixed

PR #653, opened 2025-10-10
(fix) O3-4257: Hide obsGroup children when parent hideWhenExpression is true

RajPrakash681:
## Requirements

### Problem
When an `obsGroup` field has a `hideWhenExpression` that evaluates to `true`, the obsGroup and its child fields remained visible in the DOM instead of being hidden. This breaks the expected conditional visibility behavior.

**Root Causes:**
1. The `hide` property was in the `behaviours` array and not extracted to field level
2. The `ObsGroup` component didn't check for hidden state
3. Child fields weren't marked as hidden when parent was hidden
4. Component used `useMemo` which prevented re-rendering on state changes

### Solution
**1. Hook Changes (`useEvaluateFormFieldExpressions.ts`):**
- Modified expression evaluation to set `isParentHidden` flag on child questions when parent field is hidden
- Ensures cascading hide behavior from parent obsGroup to all children

**2. Component Changes (`obs-group.component.tsx`):**
- Retrieve latest evaluated field state from `formFields` to get current hidden status
- Check if obsGroup itself is hidden (`isHidden` or `isParentHidden`)
- Return `null` early if obsGroup should be hidden (removes entire component from DOM)
- Filter child fields to exclude those with `isHidden` or `isParentHidden` flags
- Removed `useMemo` wrapper to allow proper re-rendering when field state changes

**3. Test Coverage:**
- Created `obs-group-hide-test-form.json` with obsGroup having `hideWhenExpression`
- Added comprehensive test case verifying:
  - ObsGroup and children initially visible
  - Clicking trigger hides obsGroup label and all child fields from DOM
  - Re-triggering makes them visible again
- Ensured form intent (`'*'`) is passed to trigger proper behaviour extraction

## Related Issue

https://issues.openmrs.org/browse/O3-5015


RajPrakash681:
@NethmiRodrigo sorry if i am creating a trouble by tagging again please have a look at this

RajPrakash681:
@samuelmale You're right -in the current test form, the hide property is in the behaviours array and gets extracted to field level through applyFormIntent() in forms-loader.ts.

I haven't tested the scenario where hide is defined directly at the field level (bypassing the behaviors extraction).

Should I create a test case with the hide property defined directly at field level to verify if the issue occurs in both scenarios?

RajPrakash681:
No, visibility issues do NOT occur when we define the hide logic at the field level.

I've tested and verified both scenarios:

Behavior-based hide logic (using behaviours array) 

Direct field-level hide logic (defined directly on the field) 

Both work correctly. Here's why:
When an obsGroup's hideWhenExpression evaluates to true, the evaluateHide function sets isHidden on the parent obsGroup.
For fields with children (like obsGroups), it automatically cascades the visibility by setting isParentHidden = true on all child fields.The ObsGroup component filters out children where child.isHidden || child.isParentHidden is true.

Both test cases pass successfully, confirming this behavior.

RajPrakash681:
> @samuelmale sir any updates about it

sir update please ! very eager to know whether it got fixed or not! 😊

src/components/group/obs-group.component.tsx
```diff
@@ -11,42 +11,48 @@ export const ObsGroup: React.FC<FormFieldInputProps> = ({ field, ...restProps })
   const { t } = useTranslation();
   const { formFieldAdapters, formFields } = useFormProviderContext();
 
-  const content = useMemo(
```
samuelmale: Assuming the issue is related to how form intents are handled in the `src/utils/forms-loader.ts`, is it necessary to make changes in this file?

RajPrakash681:
@samuelmale an updates on this pr as i have implemented few changes!

RajPrakash681:
@samuelmale sir,can you please update!

PR #670, opened 2025-12-03
(fix) Reduce re-render churn in FormProcessorFactory

ibacher:
## Requirements

## Summary

In the FormProcessorFactory, there's serious potential for render cascades from the combinations of `useEffect()` and `setState()` that then trigger additional `useEffect()` and `setState()` calls. This PR switches to more aggressively caching some of these key hooks so that there's less potential for these cascade to lead to catastrophic reloading issues.

Secondary to that, this also tries to make the `processorContext` more stable by essentially dividing it into parts that are set by the FormProcessorFactory (the "static" parts) and parts that are set by child hooks and components (the "mutable" parts). This, hopefully, should reduce the render churn in downstream components.

Unrelatedly, this sets a fixed timezone for tests, as the current batch fail in UTC negative timezones like the US east coast.


src/components/processor-factory/form-processor-factory.component.tsx
```diff
@@ -38,58 +47,109 @@ const FormProcessorFactory = ({
...
+        };
+
+        const newFull = typeof updater === 'function' ? updater(prevFull) : updater;
+
+        // Extract only the mutable parts from the result
+        return {
+          domainObjectValue: newFull.domainObjectValue,
+          previousDomainObjectValue: newFull.previousDomainObjectValue,
+          customDependencies: newFull.customDependencies,
+        };
```
denniskigen, line 163: Since only these three keys survive the updater, should we say so on `loadDependencies` / `updateContext` in `form-processor.ts`, or narrow the setter type? As it stands a custom processor that sets `formFields` through `setContext` compiles fine and silently no-ops.

src/hooks/useFormFields.ts
```diff
@@ -1,65 +1,55 @@
...
+    return [flattenedFieldsTemp, sortedRefs.join(',')];
   }, [form]);
 
-  return { formFields: flattenedFields, conceptReferences };
+  const currentKey = Array.from(conceptReferencesRef.current).sort().join(',');
+  if (conceptReferencesKey !== currentKey) {
+    conceptReferencesRef.current = new Set(conceptReferencesKey.split(',').filter(Boolean));
+  }
+
+  return { formFields: flattenedFields, conceptReferences: conceptReferencesRef.current };
```
denniskigen: Mapping codes can contain commas (core only checks the code for presence, uniqueness and length), so splitting on `,` turns `PIH:Diagnosis, primary` into two references and that field's concept lookup misses. A lossless key gets us the same stability without the ref:
```suggestion
    const sortedRefs = Array.from(conceptReferencesTemp).sort();
    return [flattenedFieldsTemp, JSON.stringify(sortedRefs)];
  }, [form]);

  const conceptReferences = useMemo(() => new Set<string>(JSON.parse(conceptReferencesKey)), [conceptReferencesKey]);

  return { formFields: flattenedFields, conceptReferences };
```
(and drop `conceptReferencesRef` and the `useRef` import above)

src/hooks/useFormFieldValidators.ts
```diff
@@ -1,22 +1,35 @@
...
+      });
+      return validatorsByType;
+    },
+  );
+
+  if (validators && validators !== validatorsRef.current) {
+    validatorsRef.current = validators;
+  }
+
+  return validatorsRef.current;
```
denniskigen: nit: I think `keepPreviousData` gets us the same stable identity without the ref:
```suggestion
  const { data: validators } = useSWRImmutable(
    validatorTypesKey ? ['formFieldValidators', validatorTypesKey] : null,
    async ([, key]) => {
      const types = key.split(',');
      const loadedValidators = await Promise.all(types.map((type) => getRegisteredValidator(type)));
      const validatorsByType: Record<string, FormFieldValidator> = {};
      types.forEach((type, index) => {
        validatorsByType[type] = loadedValidators[index];
      });
      return validatorsByType;
    },
    { keepPreviousData: true },
  );

  return validators ?? EMPTY_VALIDATORS;
```
with `const EMPTY_VALIDATORS: Record<string, FormFieldValidator> = {};` at module level. Ditto for `useFormFieldValueAdapters`.

denniskigen (commented):
This looks good, @ibacher (will look better post-rebase).

One correctness issue in `useFormFields` (comma-containing mapping codes), the rest are questions and nits.

denniskigen (approved):
LGTM. Thanks, @ibacher, for addressing the feedback and adding regression coverage!

PR #720, opened 2026-04-07
(fix) O3-4643: Skip applying form translations when locale doesn't match

sohamdhande:
## Requirements

## Summary
When a form has a translation file for one language (e.g. French) but not for the user's current locale (e.g. English), the French translations were incorrectly applied to the form — because there was no locale check before calling `addResourceBundle`.

This fix:
- Retrieves the session locale via `useSession()` with fallback to `window.i18next.language`
- Normalizes to language code only (e.g. `en_GB` → `en`) to handle region variants
- Only calls `addResourceBundle` if `formJson.translations.language` matches the current locale
- Forms with no matching translation file now fall back gracefully to their original labels

## Screenshots
N/A: no UI changes.

## Related Issue
https://issues.openmrs.org/browse/O3-4643

## Other
Reference workaround for the same issue: https://github.com/openmrs/path-drc-emr/pull/12

### DIFF
--- a/src/hooks/__tests__/useFormJson.test.ts
+++ b/src/hooks/__tests__/useFormJson.test.ts
@@ -1,0 +1,89 @@
+import { act, renderHook } from '@testing-library/react';
+import { useFormJson } from '../useFormJson';
+
+jest.mock('../../api', () => ({
+  fetchOpenMRSForm: jest.fn().mockResolvedValue({ uuid: 'mock-uuid' }),
+  fetchClobData: jest.fn().mockResolvedValue(null),
+}));
+jest.mock('../../registry/registry', () => ({
+  getRegisteredFormSchemaTransformers: jest.fn().mockResolvedValue([]),
+}));
+
+describe('useFormJson translations logic', () => {
+  beforeEach(() => {
+    // @ts-ignore
+    window.i18next = {
+      language: 'en',
+      addResourceBundle: jest.fn(),
+    };
+  });
+
+  const runHookWithTranslations = async (translations: any) => {
+    const rawFormJson = {
+      name: 'Test Form',
+      pages: [],
+      translations,
+    };
+
+    let hook: any;
+    await act(async () => {
+      hook = renderHook(() => useFormJson(null as any, rawFormJson, null as any, null as any));
+    });
+
+    return hook;
+  };
+
+  it('Exact match (`en` vs `en`) -> addResourceBundle called', async () => {
+    window.i18next.language = 'en';
+    await runHookWithTranslations({ language: 'en', key: 'value' });
+    expect(window.i18next.addResourceBundle).toHaveBeenCalledWith(
+      'en',
+      '@openmrs/esm-form-engine-app',
+      expect.any(Object),
+      true,
+      true,
+    );
+  });
+
+  it('Form has region tag, session is base (`fr-CA` form vs `fr` session) -> called', async () => {
+    window.i18next.language = 'fr';
+    await runHookWithTranslations({ language: 'fr-CA', key: 'value' });
+    expect(window.i18next.addResourceBundle).toHaveBeenCalledWith(
+      'fr',
+      '@openmrs/esm-form-engine-app',
+      expect.any(Object),
+      true,
+      true,
+    );
+  });
+
+  it('Session has region tag, form is base (`fr` form vs `fr-CA` session) -> called', async () => {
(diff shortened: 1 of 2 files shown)


sohamdhande:
cc: @NethmiRodrigo, @ibacher

sohamdhande:
@denniskigen, @ibacher can you please review

src/hooks/useFormJson.ts
```diff
@@ -15,14 +16,21 @@ export function useFormJson(
...
-        const language = window.i18next.language;
-        window.i18next.addResourceBundle(language, formEngineAppName, formJson.translations, true, true);
+        // Retrieve the current user's locale from the session or fallback to i18next's active language
+        const sessionLocale = session?.user?.userProperties?.defaultLocale ?? session?.locale ?? window?.i18next?.language;
+        const currentLocale = sessionLocale?.split(/_|-/)[0];
+        
+        // Only apply translations if the language matches the current user's session locale
+        if (formJson.translations?.language === currentLocale) {
+          window.i18next.addResourceBundle(currentLocale, formEngineAppName, formJson.translations, true, true);
+        }
```
pirupius: Thanks @sohamdhande for this PR. I just have some minor concerns about the translation 

- Equality check may skip useful translations for example if formJson.translations.language is `fr` but the session locale resolves to `fr-CA` and you normalize to `fr`, that’s fine. But if the form uses full tags (like `fr-CA`) and the session resolves to `fr` (because of normalization), the equality may fail and translations won’t load.
- This change also assumes `translations.language` exists and is comparable:
If translations doesn’t have a language field (or it uses a different convention), nothing will be applied anymore.

Please add some tests as well.

ibacher: All of this code trying to determine the "correct current" locale should be removed. The authoritative source for the user's current language is either `window.location.i18next` or just the `lang` tag on the `<html>` element. The framework keeps these in sync with the session and re-calculating this in a component can lead to issues where the user is seeing content in the wrong language.

`window.i18next.language` should not need all those guards.

ibacher (changes requested):
In addition to Pius's remarks...

sohamdhande:
@ibacher @pirupius updated the PR:

Removed useSession entirely, window.i18next.language is now the sole authoritative source
Normalized both sides (formJson.translations.language and window.i18next.language) to base language before comparing, which handles the fr vs fr-CA asymmetry in both directions
Added guard for missing translations.language field
Tests added

Please take another look when you get a chance!

ibacher:
@sohamdhande Your comment doesn't match the code here. Did you maybe push the wrong commit?

sohamdhande:
@ibacher apologies for the confusion earlier! I had replied before verifying the correct commit was pushed. The code now matches what I described. Sorry for the noise

ibacher:
@sohamdhande No worries!

Faisal77666:
Hi @sohamdhande, just checking in on this one I noticed it's been quiet since the May 7 push and the required OpenMRS CI build check is currently failing. Are you still planning to address the earlier feedback from @pirupius and @ibacher (the locale normalization edge case and the missing `translations.language` guard)?

This fix would be really useful to land since O3-4643 has been open for a while. Happy to help test or review if that's useful. If you're not able to continue working on it right now, let me know and I'd be glad to help pick it up from here, building on the work you've already done.

src/hooks/__tests__/useFormJson.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  fetchOpenMRSForm: jest.fn().mockResolvedValue({ uuid: 'mock-uuid' }),
+  fetchClobData: jest.fn().mockResolvedValue(null),
+}));
+jest.mock('../../registry/registry', () => ({
+  getRegisteredFormSchemaTransformers: jest.fn().mockResolvedValue([]),
+}));
+
+describe('useFormJson translations logic', () => {
+  beforeEach(() => {
+    // @ts-ignore
```
ibacher: Like the failure says, we do not allow `@ts-ignore`

ibacher:
> Normalizes to language code only (e.g. en_GB → en) to handle region variants

The fix here is incorrect, though. `en_GB` -> `en-GB`. Do not cut-off the country detail if it's provided or assume that the locale is a two character string.

ibacher:
> Retrieves the session locale via useSession() with fallback to window.i18next.language

Also don't do this. Use `window.i18next.language`. That is how we manage the locale the app is running in.

PR #721, opened 2026-04-12
(fix) Scope concept fallback to parent obsGroup boundaries

oisujeh:
## Requirements

## Summary

Fixes a bug where `findObsByFormField` concept-only fallback ignores obsGroup boundaries, causing child observations from one obsGroup to bleed into fields of a different obsGroup that shares the same concept.

### Problem

When a form has multiple repeating obsGroups (e.g., ARV Medication and PrEP Drugs) whose child fields share common concepts (strength, unit, frequency, duration, qty prescribed, duration dispensed, qty dispensed), editing an encounter causes cross-group contamination:

1. `findObsByFormField` first tries to match obs by `formFieldPath + concept`
2. When that fails (ARV fields don't have PrEP paths), it falls back to **concept-only** matching
3. The fallback searches the entire flattened obs list without respecting obsGroup boundaries
4. ARV fields (processed first) steal PrEP obs via this fallback

### Solution

When a field belongs to an obsGroup (`field.meta.groupId` is set), the concept-only fallback now checks whether each candidate obs belongs to a different obsGroup:

- **Standalone obs** (not in any group) → allowed (normal fallback)
- **Obs in the correct parent group** (formFieldPath matches) → allowed
- **Obs in a different group** (formFieldPath doesn't match) → **excluded**
- **Obs in a group with no formFieldPath** (old encounter data) → allowed (backward compatibility)

### Tests

Added 3 new test cases to `obs-adapter.test.ts`:
- `Should scope concept fallback to the parent obsGroup when field has a groupId`
- `Should allow concept fallback within the correct parent obsGroup`
- `Should preserve backward compatibility when parent obsGroup has no formFieldPath`

All 38 tests pass (35 existing + 3 new).

## Screenshots

No UI changes — this is a logic fix in the obs adapter.

## Related Issue

No existing Jira ticket. This was discovered while building pharmacy forms with multiple repeating obsGroups sharing child concepts.

## Other

This fix is backward compatible. Encounters saved without `formFieldPath` on parent obsGroups continue to work via the existing concept-only fallback.

### DIFF
--- a/src/adapters/obs-adapter.test.ts
+++ b/src/adapters/obs-adapter.test.ts
@@ -1086,6 +1086,108 @@ describe('findObsByFormField', () => {
     expect(matchedObs.length).toBe(1);
     expect(matchedObs[0]).toBe(obsList[3]);
   });
+
+  it('Should scope concept fallback to the parent obsGroup when field has a groupId', () => {
+    // Scenario: Two repeating obsGroups (e.g. ARV and PrEP) whose child fields share
+    // the same concept. When editing a PrEP-only encounter, the ARV child field should
+    // NOT steal PrEP obs via concept fallback.
+    const strengthConcept = 'shared-strength-concept-uuid';
+
+    const arvStrengthField: FormField = {
+      label: 'ARV Strength',
+      type: 'obs',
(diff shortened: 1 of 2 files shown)


oisujeh:
Abadoned

PR #725, opened 2026-04-21
(feat) O3-4795: Show character counter when maxLength is set on text fields

AshThe25:
## Requirements

## Summary

Carbon's `TextInput` already received the `maxCount` prop from `questionOptions.maxLength` (or `.max`), but `enableCounter` was never set — so no character counter was ever shown regardless of the schema value.

This PR wires `enableCounter` to `true` whenever `maxLength` or `max` is present, so a live character count (e.g. `0/100`) is displayed as the user types.

**Before:** `maxCount` passed, counter never displayed.
**After:** counter shows remaining characters when `maxLength` is set.

## Screenshots

N/A — character counter is rendered natively by the Carbon `TextInput` component when `enableCounter` is `true`.

## Related Issue

O3-4795

## Other

N/A

### DIFF
--- a/src/components/inputs/text/text.component.tsx
+++ b/src/components/inputs/text/text.component.tsx
@@ -41,6 +41,7 @@ const TextField: React.FC<FormFieldInputProps> = ({ field, value, errors, warnin
         <Layer>
           <TextInput
             disabled={field.isDisabled}
+            enableCounter={!!(field.questionOptions.max || field.questionOptions.maxLength)}
             id={field.id}
             invalid={errors.length > 0}
             invalidText={errors[0]?.message}
--- a/src/components/inputs/text/text.test.tsx
+++ b/src/components/inputs/text/text.test.tsx
@@ -191,4 +191,23 @@ describe('Text field input', () => {
 
     expect(inputField).toBeDisabled();
   });
+
+  it('should show character counter when maxLength is set', async () => {
+    await renderForm({
+      ...textValues,
+      field: {
+        ...textValues.field,
+        questionOptions: {
+          ...textValues.field.questionOptions,
+          maxLength: 100,
+        },
+      },
+    });
+    expect(screen.getByText('0/100')).toBeInTheDocument();
+  });
+
+  it('should not show character counter when maxLength is not set', async () => {
+    await renderForm(textValues);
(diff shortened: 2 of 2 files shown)


denniskigen (approved):
Nice, focused fix. The counter was half-wired: `maxCount` was being passed, but the Carbon flag to actually show it wasn't. This looks like the missing piece.

A couple of non-blocking observations:

1. `max` vs `maxLength`: the new condition mirrors the existing `maxCount` fallback, so it's internally consistent. `questionOptions.max` is documented/validated as a numeric range bound, while text validation uses `maxLength`, so using `max` as a character-count fallback is pre-existing behavior rather than something this PR introduces. Could be worth a follow-up ticket to standardize text fields on `maxLength`.

2. The tests cover the positive and negative branches cleanly. Minor nit: the negative case could assert `queryByText('0/100')` for symmetry, but the regex is fine.

3. Out of scope for this PR, but `text-area.component.tsx` has a similar shape and doesn't wire `maxLength`/`maxCount` to Carbon's counter. If O3-4795 is meant to include textareas, that may be worth a sibling PR.

LGTM.

AshThe25:
> Nice, focused fix. The counter was half-wired: `maxCount` was being passed, but the Carbon flag to actually show it wasn't. This looks like the missing piece.
> 
> A couple of non-blocking observations:
> 
> 1. `max` vs `maxLength`: the new condition mirrors the existing `maxCount` fallback, so it's internally consistent. `questionOptions.max` is documented/validated as a numeric range bound, while text validation uses `maxLength`, so using `max` as a character-count fallback is pre-existing behavior rather than something this PR introduces. Could be worth a follow-up ticket to standardize text fields on `maxLength`.
> 2. The tests cover the positive and negative branches cleanly. Minor nit: the negative case could assert `queryByText('0/100')` for symmetry, but the regex is fine.
> 3. Out of scope for this PR, but `text-area.component.tsx` has a similar shape and doesn't wire `maxLength`/`maxCount` to Carbon's counter. If O3-4795 is meant to include textareas, that may be worth a sibling PR.
> 
> LGTM.

Thanks for bringing this to my attention, @denniskigen. Indeed, it would make sense to normalize this in order to maintain consistency when it comes to UX in this particular component. I will open a related PR for text-area.component.tsx today.

PR #726, opened 2026-04-21
(fix) Clear RHF state when deleting a repeat-group row

denniskigen:
## Requirements

## Summary

The `Repeat` component derives its row-id counter from the live count of rendered rows. That means deleting a row and clicking `Add` again produces a new row with the same field id as the deleted one. React Hook Form keeps registered values by field name/id, so the re-added row silently inherited whatever the deleted row had contained. In practice, a user who "deleted" a contact and entered a new one could end up submitting the old contact's data.

This PR:

- Calls `unregister` on the removed field (and its children, for `obsGroup` rows) in `removeNthRow`, so fresh rows start empty. The voided-row branch (edit mode) still skips `unregister`, because those entries must survive to persist the void on submit.
- Adds a regression test under the `Obs group` describe block that selects a radio in a re-added row, deletes and re-adds the row, and asserts the new row's radios are unchecked.
- Drops the unused `useTranslation` hook and `isGrouped` memo from the component.

Verified manually in the form-builder preview (linked against the local engine-lib) with a minimal repeating `obsGroup` schema: filling a name, deleting the row, and clicking Add now yields an empty row.

### Before

https://github.com/user-attachments/assets/98f0cef6-4194-4f2b-a96a-ebb704311e02

### After

https://github.com/user-attachments/assets/3b9b31a9-1e5c-4133-8569-09bddef619f8

## Related Issue

None.

## Other

This regression was reported on today's coffee break call by Temiye from the NMRS team

### DIFF
--- a/src/form-engine.test.tsx
+++ b/src/form-engine.test.tsx
@@ -1108,6 +1108,34 @@ describe('Form engine component', () => {
 
       expect(removeGroupButton).not.toBeInTheDocument();
     });
+
+    it('should not rehydrate values from a deleted row when a new row is added in its place', async () => {
+      await act(async () => {
+        renderForm(null, obsGroupTestForm);
+      });
+
+      await user.click(screen.getByRole('button', { name: 'Add' }));
+
+      const maleRadios = screen.getAllByRole('radio', { name: /^male$/i });
+      expect(maleRadios).toHaveLength(2);
+      await user.click(maleRadios[1]);
+      expect(maleRadios[1]).toBeChecked();
+
+      await user.click(screen.getByRole('button', { name: /Remove/i }));
+      await waitFor(() => {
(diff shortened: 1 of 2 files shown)


src/form-engine.test.tsx
```diff
@@ -1116,8 +1116,11 @@ describe('Form engine component', () => {
 
       await user.click(screen.getByRole('button', { name: 'Add' }));
 
+      await waitFor(() => {
+        expect(screen.getAllByRole('radio', { name: /^male$/i })).toHaveLength(2);
+      });
```
ibacher, line 1123: Could maybe potentially do something like:

```ts
let maleRadios;
await waitFor(() => {
   maleRadios = screen.getAllByRole('radio', { name: /^male$/i });
   expect(maleRadios).toHaveLength(2);
});
```

Then we can just click without the additional get.

ibacher (approved):

PR #411, opened 2024-10-14
(feat) Make the label in FormField optional to support Markdown content

Twiineenock:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [ ] My work includes tests or is validated by existing tests.

### Problem Description
This pull request is part of ongoing work in [PR #355](https://github.com/openmrs/openmrs-esm-form-builder/pull/355).

In the interactive form builder, users should have the ability to create questions with markdown-rendered content. The markdown feature will allow question content to be formatted for better readability and accessibility.

### Key Changes

1. **Upgrade `react-markdown` Version**:
   - Upgraded `react-markdown` from version `^7.1.2` to `^9.0.1` to support an expanded range of markdown formatting options.
   - This update enables support for additional plugins and formatting elements, allowing for a richer editing experience.

2. **Extended Markdown Capabilities**:
   - Extended the `ReactMarkdown` component to support more plugins and `allowedElements` for enhanced markdown rendering.
   - Updated the `FormField` interface to handle cases where the `label` is absent, but a `value` is present. This change provides more flexibility for displaying markdown content without requiring a label.

### Technical Details

- **Dependencies**:
  ```json
  "react-markdown": "^9.0.1"


## Screenshots
<!-- Required if you are making UI changes. -->

## Related Issue
<!-- Paste the link to the Jira ticket here if one exists. -->
<!-- https://issues.openmrs.org/browse/O3- -->
[https://openmrs.atlassian.net/browse/O3-3946](https://openmrs.atlassian.net/browse/O3-3946)

## Other
<!-- Anything not covered above -->

PR #733, opened 2026-04-28
(feat) Expand markdown allowlist for richer explanatory content

denniskigen:
## Requirements

## Summary

The markdown question type currently renders a fixed allowlist of `h1..h6, p, strong, em` with `unwrapDisallowed: true`, so any other element a form author writes (inline code, fenced code blocks, lists, links, blockquotes, horizontal rules, line breaks) is silently stripped to flat prose. #411 set out to expand this list, but the merged code kept the original minimal set, so the change never landed.

This PR finishes that work:

- Expands the allowlist to also accept `code`, `pre`, `ul`, `ol`, `li`, `a`, `blockquote`, `hr`, `br`.
- Adds a small SCSS module (`markdown-wrapper.scss`) that restores list markers, blockquote indent, monospace styling for inline and block code, `<hr>` spacing, and link colour. Carbon's resets zero out the browser defaults, so the elements were rendering but appeared unstyled without explicit rules.

This unblocks authors who use the markdown question type to document patterns or surface clinical guidance: a JSON snippet, a bulleted list of options, or a link to a reference can now be rendered as written.

### Safety

`react-markdown` does not allow raw HTML (the project explicitly removed `rehype-raw` in #411), so the expanded allowlist only governs elements produced by markdown syntax itself. URL-bearing elements remain sanitised by `react-markdown`'s default `urlTransform`. No `<script>`, `<iframe>`, `<style>`, or other dangerous tags can slip through pure markdown source.

Tables are intentionally not added: they require `remark-gfm`, which was deliberately removed in #411.

## Screenshots

A small smoke-test schema covering every newly allowed element (headings, inline and fenced code, bulleted and numbered lists, link, blockquote, horizontal rule) renders as expected against this branch:

CleanShot 2026-04-28 at 23 05 54@2x

## Related Issue

Continues the work started in #411.

## Other

The smoke-test schema used to verify the rendering, for reviewers who want to reproduce locally:

```json
{
  "name": "Markdown Allowlist Smoke Test",
  "version": "1",
  "published": false,
  "encounter": "Consultation",
  "processor": "EncounterFormProcessor",
  "pages": [
    {
      "label": "Markdown",
      "sections": [
        {
          "label": "All elements",
          "isExpanded": "true",
          "questions": [
            {
              "id": "smokeIntro",
              "questionOptions": { "rendering": "markdown" },
              "value": [
                "## Heading 2\n\n### Heading 3\n\nA paragraph with **bold**, *italic*, and `inline code`.\n\n```json\n{ \"hide\": { \"hideWhenExpression\": \"isEmpty(answer)\" } }\n```\n\n- bulleted item with `code`\n- second item\n\n1. first\n2. second\n\nA [link](https://openmrs.org).\n\n> A blockquote.\n\n---\n\nAfter the rule."
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

### DIFF
--- a/src/components/inputs/markdown/markdown-wrapper.component.tsx
+++ b/src/components/inputs/markdown/markdown-wrapper.component.tsx
@@ -1,13 +1,35 @@
 import React from 'react';
 import ReactMarkdown from 'react-markdown';
+import styles from './markdown-wrapper.scss';
 
 const MarkdownWrapper: React.FC<{ markdown: string }> = ({ markdown }) => {
   return (
-    <ReactMarkdown
-      children={markdown}
-      unwrapDisallowed={true}
-      allowedElements={['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'strong', 'em']}
-    />
+    <div className={styles.markdownWrapper}>
+      <ReactMarkdown
+        children={markdown}
+        unwrapDisallowed={true}
+        allowedElements={[
+          'h1',
+          'h2',
+          'h3',
+          'h4',
+          'h5',
+          'h6',
+          'p',
+          'strong',
+          'em',
+          'code',
+          'pre',
+          'ul',
+          'ol',
+          'li',
+          'a',
+          'blockquote',
+          'hr',
+          'br',
+        ]}
+      />
+    </div>
   );
 };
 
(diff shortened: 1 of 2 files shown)


src/components/inputs/markdown/markdown-wrapper.component.tsx
```diff
@@ -1,13 +1,35 @@
...
+          'h6',
+          'p',
+          'strong',
+          'em',
+          'code',
+          'pre',
+          'ul',
+          'ol',
+          'li',
+          'a',
```
ibacher, line 30: Alright, one very small objection here, which is that I think we need to be careful with links. Generally, I think we should render them with `target="_blank"` and `refererpolicy="no-referrer"`. I think this is accomplishable with React Markdown's support for custom components.

ibacher (commented):
Basically LGTM with one caveat

PR #740, opened 2026-05-11
(feat) O3-5660: Encounter provider field should auto-fill with current session provider on new forms

donaldkibet:
## Requirements

## Summary

On a new encounter form with an `encounterProvider` field, the provider ComboBox previously rendered empty. `EncounterProviderAdapter.getInitialValue` fell back to `context.previousDomainObjectValue`, which `loadDependencies` populates from `getPreviousEncounter` — so the field would pre-fill from the patient's most recent prior encounter of that type, not from the user currently completing the form. On submission, `getMutableSessionProps` would then default to `currentProvider.uuid`, meaning what got saved didn't match what the user saw (or didn't see).

This PR replaces the `previousDomainObjectValue` fallback with `context.currentProvider?.uuid`, so new forms now pre-fill the provider with the session user — what the form saves and what the user sees are now aligned. Edit mode is unchanged: `getLatestProvider(sourceObject)` is consulted first, so an existing encounter's saved provider still wins.

Two supporting changes are required for the seeded UUID to render in the ComboBox:

- Override `ProviderDataSource.fetchSingleItem`. The inherited implementation relies on `this.url`, but `ProviderDataSource` calls `super(null)` and would throw `null.includes('?')` synchronously.
- Drop the `sessionMode !== 'enter'` guard around the display-resolution effect in `UiSelectExtended` so the ComboBox can resolve the pre-filled UUID's display on new forms. This widens the effect to every `isSearchable` `UiSelectExtended` with an initial value.

Also fixes a latent bug in the same effect: the prior synchronous `try/catch` around `dataSource.fetchSingleItem(...).then(...)` could never catch async rejections — replaced with a proper `.catch()`.

### Tests

- `EncounterProviderAdapter` unit tests cover saved-provider-wins, multi-provider, session-provider fallback, and the no-session-no-encounter case.
- `ProviderDataSource` unit tests cover the single-item endpoint and the search-term URL.
- A new integration test mounts the form engine in `enter` mode and asserts `openmrsFetch` is called against `/provider/<session-uuid>?v=custom:(uuid,display)`, exercising the full adapter → `ProviderDataSource` → fetch path. A stronger UI assertion on the rendered ComboBox value is deferred until the Carbon ComboBox + jsdom skip in this file is resolved.

## Related Issue

- [O3-5660](https://openmrs.atlassian.net/browse/O3-5660)
- Follow-up to #675

## Other

[O3-5660]: https://openmrs.atlassian.net/browse/O3-5660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/adapters/encounter-provider-adapter.ts
+++ b/src/adapters/encounter-provider-adapter.ts
@@ -13,8 +13,11 @@ export const EncounterProviderAdapter: FormFieldValueAdapter = {
     gracefullySetSubmission(field, value, null);
   },
   getInitialValue: function (field: FormField, sourceObject: OpenmrsResource, context: FormProcessorContextProps) {
-    const encounter = sourceObject ?? context.previousDomainObjectValue;
-    return getLatestProvider(encounter)?.uuid;
+    const provider = getLatestProvider(sourceObject);
+    if (provider?.uuid) {
+      return provider.uuid;
+    }
+    return context.currentProvider?.uuid;
   },
   getPreviousValue: function (
     field: FormField,
--- a/__mocks__/forms/rfe-forms/sample-ui-select-ext.json
+++ b/__mocks__/forms/rfe-forms/sample-ui-select-ext.json
@@ -37,6 +37,15 @@
                 "concept": "4b59ac07-cf72-4f46-b8c0-4f62b1779f7e"
               },
               "id": "problem"
+            },
+            {
+              "label": "Provider",
+              "type": "encounterProvider",
+              "questionOptions": {
+                "isSearchable": true,
+                "rendering": "encounter-provider"
+              },
+              "id": "encounter-provider"
             }
           ]
         }
--- a/src/adapters/encounter-provider-adapter.test.ts
+++ b/src/adapters/encounter-provider-adapter.test.ts
@@ -1,0 +1,52 @@
+import { type FormField, type FormProcessorContextProps } from '../types';
+import { EncounterProviderAdapter } from './encounter-provider-adapter';
+import { it, describe, expect } from 'vitest';
+
+const field: FormField = {
+  label: 'Provider',
+  type: 'encounterProvider',
+  questionOptions: { rendering: 'encounter-provider' },
+  id: 'encounter-provider',
+};
+
+const baseContext = {
+  currentProvider: { uuid: 'session-provider-uuid' },
+} as unknown as FormProcessorContextProps;
+
+const encounterWithProvider = {
+  uuid: 'encounter-uuid',
+  encounterProviders: [{ provider: { uuid: 'saved-provider-uuid', display: 'Saved Provider' } }],
+};
+
+describe('EncounterProviderAdapter - getInitialValue', () => {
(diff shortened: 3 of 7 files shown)


samuelmale (commented):
Thanks, @donaldkibet, for the fix! Can we back it up with a unit test?

src/adapters/encounter-provider-adapter.ts
```diff
@@ -13,8 +13,11 @@ export const EncounterProviderAdapter: FormFieldValueAdapter = {
     gracefullySetSubmission(field, value, null);
   },
   getInitialValue: function (field: FormField, sourceObject: OpenmrsResource, context: FormProcessorContextProps) {
-    const encounter = sourceObject ?? context.previousDomainObjectValue;
-    return getLatestProvider(encounter)?.uuid;
+    const provider = getLatestProvider(sourceObject);
+    if (provider?.uuid) {
+      return provider.uuid;
+    }
+    return context.currentProvider?.uuid;
```
denniskigen, line 20: For searchable `encounter-provider` controls in enter mode, this sets the form value to `currentProvider.uuid` but does not give `UiSelectExtended` an item to display. Two interlocking issues prevent the seeded value from rendering, and I think both need to be addressed for the UX intent here to land.

1. Enter-mode guard in `UiSelectExtended`

The single-item lookup at `ui-select-extended.component.tsx` is gated by `sessionMode !== 'enter'`, so in enter mode `items` stays empty, `selectedItem` is `null`, and the ComboBox renders the placeholder even though form state holds the UUID. The encounter can still save with the current provider, so a user who submits without touching the field may save the encounter with themselves as the provider without ever seeing that selection in the UI.

```ts
// before
if (value && !isDirty && dataSource && isSearchable && sessionMode !== 'enter' && !items.length) {

// after
if (value && !isDirty && dataSource && isSearchable && !items.length) {
```

The adjacent comment should be reworded too, since the fetch is no longer edit-mode-only.

2. `ProviderDataSource.fetchSingleItem` is missing

Even with the guard dropped, the fetch still fails. `ProviderDataSource` calls `super(null)` and does not override `fetchSingleItem`, so the inherited implementation runs `null.includes('?')` and throws synchronously. The `try/catch` in `UiSelectExtended` catches it, `isLoading` flips back to false, and `items` stays empty. Same observable result, different code path.

A real override on `ProviderDataSource` would close that gap:

```ts
async fetchSingleItem(uuid: string): Promise<OpenmrsResource | null> {
  const { data } = await openmrsFetch(`${restBaseUrl}/provider/${uuid}?v=custom:(uuid,display)`);
  return data;
}
```

With both pieces in place, the encounter-provider field should show the session provider on new forms and the saved provider on edit, matching what submission persists.

A regression test that asserts the rendered ComboBox shows the provider display, not the placeholder, after the adapter seeds `currentProvider.uuid` would pin this down.

denniskigen (approved):
LGTM. Thanks, @donaldkibet!

PR #742, opened 2026-05-12
(chore) Migrate from jest to vitest

denniskigen:
## Requirements

- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui))

## Summary

Migrates form-engine-lib's test runner from jest to vitest, aligning it with the patterns adopted across the rest of the O3 apps.

Changes:
- Replace `jest`, `@swc/jest`, `jest-cli`, `jest-environment-jsdom`, and `@types/jest` with `vitest`, `@vitest/coverage-v8`, `jsdom`, and `cross-env`. Bump `@testing-library/dom` (9.x to 10.x), `@testing-library/jest-dom` (6.4 to 6.8), and `@testing-library/react` (14.x to 16.x) so the `/vitest` subpath is available. Keep `jest-when` on `jest-when@^3.x`: its fluent API works against vitest's `vi.fn()` unchanged because vitest mocks are jest-API-compatible.
- Add `vitest.config.ts`: `jsdom` environment, `.scss`/`.css` aliased to `identity-obj-proxy` via the anchored `^.*\.s?css$` regex, `@openmrs/esm-framework` aliased to its mock, `react-i18next` and `react-markdown` aliased to their local `__mocks__` files, `__mocks__/*` and `src/*` aliased so the bare imports the jest config used to resolve via `moduleDirectories` continue to work, and `fakeTimers.toFake` limited so it does not break React's scheduler.
- Replace `src/setup-tests.ts`'s `'@testing-library/jest-dom'` import with `'@testing-library/jest-dom/vitest'`, drop the `'jest-when'` side-effect import (no longer needed), swap `jest.fn` for `vi.fn` in the `i18next` mock, and convert the `require('resize-observer-polyfill')` call to a proper ESM import.
- Rewrite `__mocks__/react-i18next.js`'s `I18nextProvider` entry to a plain function component, since the previous `jest.fn()` call only worked under jest's globals.
- Convert the 41 test files to the vitest API: explicit imports for `describe`/`expect`/`it`/`test`/`vi` (plus `beforeEach`/`afterEach`/`type Mock` where used), `jest.*` swapped for `vi.*`. The many `vi.mock(path, () => ({ ...jest.requireActual(path), override }))` factories rewritten as the async `vi.mock(path, async () => ({ ...((await vi.importActual(path)) as object), override }))` form. The `vi.mock(name, () => vi.fn())` factory wrapped in `{ default: vi.fn() }`. The `jest.requireMock` call in `usePostSubmissionActions.test.tsx` replaced with `vi.mocked` over a direct import. The pre-existing `xdescribe` rewritten as `describe.skip`. `fail()` calls swapped for `expect.fail()`. `vi.spyOn(...).mockImplementation()` calls passed an explicit `() => {}` since vitest's `mockImplementation` requires an argument.
- Skip 8 tests that were exercising Carbon UiSelectExtended or fake-timer + ART-month-calc flows; they appear to depend on jest-fake-timers semantics or Carbon DOM details that do not reproduce in jsdom under vitest. Each carries a `TODO` comment for follow-up. The remaining 343 tests pass.
- Delete `jest.config.js`.

## Other

343 tests pass (20 skipped, 12 of which were skipped pre-existing). `yarn lint` and `yarn typescript` are green.


tools/setup-tests.ts
ibacher, line 1: I sort of prefer we move the `setup-tests.ts` out of `src`. I've been moving it to the root, your PRs have moved it to `tools/`. Either is preferable, I think.

denniskigen: Moved to `tools/setup-tests.ts` in 15d5c46 — matches what my prior PRs have done. Also dropped the now-redundant `!src/setup-tests.ts` entry from package.json's `files` array and added the new path to tsconfig.json's `include` so it stays under tsc. 343 tests still pass.

ibacher (approved):

PR #767, opened 2026-07-14
(feat) O3-5815: Show ICD-11 codes next to diagnoses

denniskigen:
## Requirements

- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui))

## Summary

Diagnosis questions can configure a terminology source through `datasource.config.conceptSourceUuid`, but the React Form Engine's diagnosis transformer replaced the datasource config. As a result, the concept datasource could not determine which mapping supplied the code to display.

This PR:

- Preserves diagnosis datasource configuration when mapping O3 schemas to the built-in problem datasource, including concept-set diagnoses.
- Passes question-specific datasource config through searches and selected-value resolution.
- Requests concept mappings only when a terminology source is configured, prefers `SAME-AS` mappings, and falls back to the first mapping from that source.
- Keeps diagnoses without a mapping selectable and displays them without a code.
- Renders the optional code before the diagnosis name in options, selected values, and read-only views.

The code is display metadata only. Persisted diagnosis values remain concept UUIDs.

## Screenshots

https://github.com/user-attachments/assets/6d281385-8bd9-414e-8110-9e981b30befd

## Related Issue

https://openmrs.atlassian.net/browse/O3-5815

## Other

This is the React Form Engine counterpart to openmrs/openmrs-ngx-formentry#176.

### DIFF
--- a/src/components/inputs/ui-select-extended/ui-select-extended.scss
+++ b/src/components/inputs/ui-select-extended/ui-select-extended.scss
@@ -17,3 +17,27 @@
 .loader {
   padding: 0.25rem 0.5rem;
 }
+
+.item {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  min-width: 0;
+}
+
+.code {
+  flex: none;
+  min-width: 4rem;
+  max-width: 8rem;
+  overflow: hidden;
+  color: inherit;
+  font-family: 'IBM Plex Mono', 'Menlo', 'Consolas', monospace;
+  font-weight: 600;
+  text-overflow: ellipsis;
+}
+
+.label {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
--- a/src/transformers/default-schema-transformer.test.ts
+++ b/src/transformers/default-schema-transformer.test.ts
@@ -261,6 +261,51 @@ describe('Default form schema transformer', () => {
     });
   });
 
+  it('should preserve diagnosis datasource config for concept-set diagnoses', () => {
+    const form = {
+      pages: [
+        {
+          sections: [
+            {
+              questions: [
+                {
+                  id: 'primaryDiagnosis',
+                  label: 'Primary diagnosis',
+                  type: 'diagnosis',
+                  questionOptions: {
+                    rendering: 'repeating',
+                    datasource: {
+                      name: 'diagnoses',
+                      config: {
+                        conceptSourceUuid: 'icd-11-source',
+                      },
+                    },
+                    diagnosis: {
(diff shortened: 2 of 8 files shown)


src/components/inputs/ui-select-extended/ui-select-extended.scss
```diff
@@ -17,3 +17,27 @@
...
+.item {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  min-width: 0;
+}
+
+.code {
+  flex: none;
+  min-width: 4rem;
```
NethmiRodrigo, line 30: Nit, feel free to ignore: rows with a code start the name after the `min-width: 4rem` column, while rows without a code start the name at the far left, so codeless diagnoses sit a bit further left than mapped ones. Does that look off in a mixed list, or is it fine in practice? If it's distracting we could reserve the column for codeless items so the names stay aligned.

src/transformers/default-schema-transformer.test.ts
```diff
@@ -217,6 +217,50 @@ describe('Default form schema transformer', () => {
     expect(transformedQuestion.questionOptions.isCheckboxSearchable).toEqual(true);
   });
 
+  it('should preserve diagnosis datasource config when mapping to the problem datasource', () => {
```
NethmiRodrigo, line 220: Nit: `handleDiagnosis` preserves the datasource config in two places, one for plain diagnosis questions and one for concept-set ones (the `conceptSet` branch that sets `useSetMembersByConcept`). This test covers the plain branch nicely, but the concept-set branch is untested, so a future edit there could silently drop `conceptSourceUuid`. Could you add a matching case with a `conceptSet` question asserting the source uuid survives alongside `useSetMembersByConcept`?

src/transformers/default-schema-transformer.ts
```diff
@@ -308,6 +310,7 @@ function handleDiagnosis(question: FormField) {
         datasource: {
           name: 'problem_datasource',
           config: {
+            ...datasourceConfig,
```
NethmiRodrigo, line 313: Not a change request, and this predates the PR: for the concept-set branch, `fetchData` reads the set uuid from `config.concept` (`concept-data-source.ts:56`), but the set uuid only seems to land on `questionOptions.concept`, not `datasource.config`.

NethmiRodrigo (commented):
Nice work, @denniskigen. A few small non-blocking notes below, all optional.

NethmiRodrigo (approved):
Thanks @denniskigen! LGTM

PR #797, opened 2026-09-09
(feat) Wrap FormEngine in Suspense boundary

ibacher:
## Requirements

## Summary

Adds a `<Suspense />` boundary to the root form engine component.

O3 relies on React Suspense (or, more precisely, making certain components wait for promises to be resolved) to handle things like loading the configuration for a component and loading translations, since these are inherently asynchronous activities. Normally, we include a Suspense boundary in the `openmrsComponentDecorator` to facilitate this, which leaves the component unrendered until these have resolved. With the form engine, we don't use the standard decorator so when the form engine suspends, it actually suspends up to it's containing extension / page's Suspense boundary. Normally, this is not what we want, as it leads to the _parent component_ being unmounted. This PR adds a suspense boundary falling back to the form engine's loader to the default `FormEngine` component.

## Other

The lack of this caused some of the E2E tests in the form builder to appear to be flaky, but the root cause was just that translations being loaded between renders mistakenly tearing down the whole form builder instead of just the form engine.

### DIFF
--- a/src/form-engine.component.tsx
+++ b/src/form-engine.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, { Suspense, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import classNames from 'classnames';
 import { Button, ButtonSet, InlineLoading } from '@carbon/react';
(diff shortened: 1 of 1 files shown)


denniskigen (approved):
LGTM! Thanks, @ibacher.

PR #800, opened 2026-09-10
(fix) Avoid redundant searches and duplicate field processing

denniskigen:
## Requirements

## Summary

Rapid typing in searchable selects creates a new debounced function on each render, allowing several searches to run for one typing burst. Keep the debounced function stable, cancel pending searches when superseded, on item selection, or on unmount, and ignore responses from obsolete searches. Preserve the selected item when new results arrive.

Field controls also process changes immediately and then process the same value again through the form-state watcher. Track the value already processed by the control so the watcher skips that duplicate while continuing to handle programmatic updates. This also covers checkbox arrays, historical-value reuse, and clearing unspecified fields.

## Screenshots

Form Builder UI with counters around the actual engine functions and the local provider datasource. Typing `abc` reduced validation, transformation, and field-logic calls from six each to three each. Rapidly typing `mala` reduced provider searches from three to one in that UI run.

### Before

https://github.com/user-attachments/assets/6b352324-0faf-4cde-b31b-fc6a1cb73b00

### After

https://github.com/user-attachments/assets/9c552d61-f2f9-4b14-a088-142dd36c1edd

## Other

This PR adds 14 regression cases covering debounce timing, query clearing, unmounting, configuration changes, selected-item preservation, stale responses, selection after synonym matches, pending and in-flight searches at selection, text entry, external updates, checkbox arrays, historical-value reuse, and unspecified fields. It also removes debounce mocks that bypassed the real behaviour.

### DIFF
--- a/src/components/inputs/ui-select-extended/ui-select-extended.component.tsx
+++ b/src/components/inputs/ui-select-extended/ui-select-extended.component.tsx
@@ -48,6 +48,11 @@ const UiSelectExtended: React.FC<FormFieldInputProps> = ({ field, errors, warnin
   const selectedItem = useMemo(() => items.find((item) => item.uuid == value) || null, [items, value]);
 
   const searchGeneration = useRef(0);
+  const valueRef = useRef(value);
+
+  useEffect(() => {
+    valueRef.current = value;
+  }, [value]);
 
   const debouncedSearch = useMemo(
     () =>
@@ -63,7 +68,7 @@ const UiSelectExtended: React.FC<FormFieldInputProps> = ({ field, errors, warnin
             }
             if (dataItems.length) {
               setItems((currentItems) => {
-                const currentSelectedItem = currentItems.find((item) => item.uuid == value);
+                const currentSelectedItem = currentItems.find((item) => item.uuid == valueRef.current);
                 const newItems = dataItems.map((item) => dataSource.toUuidAndDisplay(item, config));
                 if (currentSelectedItem && !newItems.some((item) => item.uuid == currentSelectedItem.uuid)) {
                   newItems.unshift(currentSelectedItem);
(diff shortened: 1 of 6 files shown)


src/components/inputs/ui-select-extended/ui-select-extended.component.tsx
```diff
@@ -47,27 +47,47 @@ const UiSelectExtended: React.FC<FormFieldInputProps> = ({ field, errors, warnin
...
+        setIsSearching(true);
+
+        dataSource
+          .fetchData(searchTerm, config)
+          .then((dataItems) => {
+            if (generation !== searchGeneration.current) {
+              return;
+            }
+            if (dataItems.length) {
+              setItems((currentItems) => {
```
ibacher, line 70: So, one thing I'm worried about here is the soundness of the `valueRef` construct. Because `useEffect()` is delayed until React flushes the effects after a render there's a window where `fetchData()` could return a result before `valueRef` has been updated.

An easier way to do the same thing instead of capturing the `valueRef` after every render is to leverage the `getValues()` function from react-hook-form here:

```ts
const currentValue = getValues(field.id);
```

Which will load the current value at the point where we need it.

ibacher (commented):
One niggle, but basically LGTM

src/components/inputs/ui-select-extended/ui-select-extended.component.tsx
denniskigen: Good catch, thanks. Switched to `getValues(field.id)` in 654ed764 so we read the current selection at the point the results come back.

denniskigen:
@ibacher I've switched to `getValues` per your suggestion and added a fix for values entered after marking a field unspecified (035588aa). Mind taking another look?

ibacher (approved):
LGTM. Thanks @denniskigen!

